PR #27, opened 2025-01-07
Add Postgres to CI environment

sujayakar:
----


nipunn1313:
https://github.com/get-convex/convex/blob/main/.github/workflows/build_local_backend.yml.oss

nipunn1313:
once you get it working, patch the change here https://github.com/get-convex/convex/blob/main/.github/workflows/build_local_backend.yml.oss and land it

PR #46, opened 2025-02-18
Update run_backend.sh

satybaldin:
#### **Summary**  
This PR improves the database configuration logic in the Convex startup script to ensure proper handling of PostgreSQL and SQLite connections.  

#### **Changes made**  
1. **Ensured correct PostgreSQL connection flag order**  
   - The official Convex Discord community mentioned that `--db postgres-v5` **must** come before the connection string.  
   - Fixed the ordering issue in the script.  

2. **Added conditional SSL disabling for PostgreSQL**  
   - Introduced logic to **conditionally** add `--do-not-require-ssl` if the user explicitly sets it.  
   - The flag is **not** included by default, ensuring security for production environments.  

3. **Fixed SQLite execution logic**  
   - SQLite should receive `"${DB_FLAGS[@]}" "$DB_SPEC"`, while PostgreSQL should only receive `"${DB_FLAGS[@]}"`.  
   - This prevents SQLite from breaking due to unnecessary parameters.  

4. **Removed `.env` file parsing**  
   - The official Convex script does not load `.env` explicitly.  
   - This aligns the behavior with the official implementation.  

#### **Remaining Issues / Suggestions**  
✅ **This script now properly handles both PostgreSQL and SQLite**, but a few improvements could be made in the future:  
- **Dynamic TLS handling for PostgreSQL** instead of forcing `--do-not-require-ssl`.  
- **Better logging for database connection failures** to help with debugging.  
- **More flexible configuration options**, such as allowing users to override database flags.  

This PR ensures that users can **properly self-host Convex with PostgreSQL and SQLite** without running into misconfiguration issues.

### DIFF
--- a/self-hosted/docker-build/run_backend.sh
+++ b/self-hosted/docker-build/run_backend.sh
@@ -1,5 +1,6 @@
 #! /bin/bash
 
+# Set data directories
 export DATA_DIR=${DATA_DIR:-/convex/data}
 export TMPDIR=${TMPDIR:-"$DATA_DIR/tmp"}
 export STORAGE_DIR=${STORAGE_DIR:-"$DATA_DIR/storage"}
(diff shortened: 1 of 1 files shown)


emmaling27:
Hey @satybaldin! Thanks for the contribution. I already updated this in https://github.com/get-convex/convex-backend/commit/62ef09aa604b0c5f873b59e0944b5e89f84b66b2, so I'm closing this PR since it does the same thing. Thanks for finding this issue! It has been built and tested, and it's available [here](https://github.com/get-convex/convex-backend/pkgs/container/convex-backend/358166327?tag=62ef09aa604b0c5f873b59e0944b5e89f84b66b2). The docs are updated with instructions on how to run with local Postgres [here](https://github.com/get-convex/convex-backend/tree/main/self-hosted#connecting-to-postgres-locally)

PR #51, opened 2025-02-28
Add second workaround to circular type inference

benjavicente:
I added a second option to solve the circular type inference in actions: typing the handler function return type. It might be simpler than typing `runQuery` and `runMutation`, specially when there are multiple queries and mutations, so I place it first.

I also made the title an h3, to be visible in the TOC, and because I think it is more of a problem of the action constructor been part of the type generation than the context. Maybe it need to be part of another page (I don't known witch could be), because this could also happen on other functions, like mutations with partial rollback 🤔


emmaling27:
@benjavicente thank you for the contribution! I merged it in here: https://github.com/get-convex/convex-backend/commit/8e98fae0eccbef33a9ed5eb7bf4c9d3f1c66b1d6

PR #52, opened 2025-03-01
Fix docs for Clerk `getAuthToken`

TheMythologist:
The current example provided is invalid. I've updated the docs to reflect the proper way to implement `getAuthToken` for Clerk


emmaling27:
@TheMythologist thank you for the contribution! I merged it in here: https://github.com/get-convex/convex-backend/commit/9955fa46efbf3136af170a7a45af1c979a9526b1

PR #53, opened 2025-03-02
Add support for Cloudflare R2 Storage

Spioune:
This PR introduces support for Cloudflare R2 by adding a new environment variable, AWS_ENDPOINT_URL, and replacing get_object_attributes with head_object.

From the [S3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3/client/get_object_attributes.html), get_object_attributes combines the functionality of HeadObject and ListParts in a single call.
In the original code get_object_attributes was used to get the file size.
Since R2 does not support get_object_attributes, I use head_object instead.

@emmaling27 I would like your help on this.
I wanted to name the env variable S3_ENDPOINT_URL but I read that AWS automatically provides AWS_ENDPOINT_URL to lambda functions.
I can see you use `aws_config::defaults(BehaviorVersion::v2024_03_28())` which is equivalent to `aws_config::from_env()`, which I assumed automatically set defaults based on env variables ? From my tests it doesn't seem to do anything.
So right now I call `.endpoint_url(endpoint_url)` explicitly.
What is the best way to make it optional ? (I don't know much about rust)
Right now it's a mandatory variable ⚠️ which is not what we want.

----

### DIFF
--- a/crates/aws_s3/src/storage.rs
+++ b/crates/aws_s3/src/storage.rs
@@ -9,9 +9,9 @@ use async_trait::async_trait;
 use aws_config::retry::RetryConfig;
 use aws_sdk_s3::{
     operation::{
-        get_object_attributes::{
-            GetObjectAttributesError,
-            GetObjectAttributesOutput,
+        head_object::{
+            HeadObjectOutput,
+            HeadObjectError,
         },
         upload_part::builders::UploadPartFluentBuilder,
     },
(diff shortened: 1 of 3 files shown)


emmaling27:
Great work! Thanks so much for adding this 🙏🏼 I'm sure lots of folks will find it useful to use R2 and save some 💸 

Good question how to add the `S3_ENDPOINT_URL` and not have it affect lambdas... I would recommend adding a wrapper function `must_s3_config_from_env` that calls the unchanged `must_config_from_env` function and adds the endpoint url.
```rust
pub fn must_s3_config_from_env() -> anyhow::Result<ConfigLoader> {
    let mut config_loader = must_config_from_env()?;
    if let Some(s3_endpoint_url) = S3_ENDPOINT_URL.clone() {
        config_loader = config_loader.endpoint_url(s3_endpoint_url);
    }
    Ok(config_loader)
}
```
And then calling this from the `S3Client`. Let me know if that makes sense!

Spioune:
Thanks for your reply.
Yes that makes sense, I will try to do that.

But don't you feel it's weird to have to explicitly `.region(region).endpoint_url(endpoint_url).credentials(credentials)` when aws_conf::from_env should do this by default already ?

emmaling27:
> Thanks for your reply. Yes that makes sense, I will try to do that.
> 
> But don't you feel it's weird to have to explicitly `.region(region).endpoint_url(endpoint_url).credentials(credentials)` when aws_conf::from_env should do this by default already ?

We add them explicitly so we can check and assert that they are there and give a better error message.

Spioune:
> > Thanks for your reply. Yes that makes sense, I will try to do that.
> > But don't you feel it's weird to have to explicitly `.region(region).endpoint_url(endpoint_url).credentials(credentials)` when aws_conf::from_env should do this by default already ?
> 
> We add them explicitly so we can check and assert that they are there and give a better error message.

Ok, then I think this is good to go now.

emmaling27:
@Spioune thank you so much! I've merged this in here: https://github.com/get-convex/convex-backend/commit/45279e3ba77515b092db819d69ed7d2e6f60c14e. I'll build a new docker image and release tomorrow.

I also added a new issue I noticed with this setup - that you can easily switch between R2 and S3 without receiving an error because the `StorageTagInitializer` stored in the database doesn't know you're using a different storage URL. This could lead to lots of errors and backend data getting in a bad state with files split between two different storages. Do you want to take a stab at it! https://github.com/get-convex/convex-backend/issues/55

Spioune:
@emmaling27 thank you for your help ! I am happy to be able to use it on my projects without needing to maintain my own fork.

Good point, I will take a look at the issue.

PR #59, opened 2025-03-12
Improve `.unique()` error message: print document `_id`s

djbalin:
Makes the error message thrown by `.unique()` more useful. The `_id`'s of the first two documents returned by the `unique()` query are printed in the error message. This is very helpful to the user in identifying and resolving incorrect data.

Feel free to change my wording in the error message - but in my opinion it should be clear that the document id's printed are *not* all the documents returned by the `unique()` query but just the first two.

I [built the Convex backend locally](https://github.com/get-convex/convex-backend/blob/ad552652f4aaa0225e977ed7f5d92a35108b85d4/BUILD.md), made this change locally, and ran the tutorial demo to verify that it works as expected.

----


npm-packages/convex/src/server/impl/query_impl.ts
```diff
@@ -326,7 +326,7 @@ export class QueryImpl implements Query<GenericTableInfo> {
       return null;
     }
     if (first_two_array.length === 2) {
-      throw new Error("unique() query returned more than one result");
+      throw new Error(`unique() query returned more than one result. First pair of results: ${first_two_array[0]._id} and ${first_two_array[1]._id}`);
```
nipunn1313, line 329:
```suggestion
      throw new Error(`unique() query returned more than one result: 
 [${first_two_array[0]._id}, ${first_two_array[1]._id}, ...]`);
```

Suggestion for a little bit more terse way of showing it.
This is a great idea!

What do you think?

nipunn1313 (approved):
looks good - I'll pull it in. Might take a sec, since we have to run some tests behind the scenes.

npm-packages/convex/src/server/impl/query_impl.ts
djbalin: Thanks, looks good to me! I have [opened a PR](https://github.com/get-convex/convex-helpers/pull/495) for basically the same problem in convex-helpers. The `unique()` function in that library just prints `"not unique"`. imo, it's confusing that the error messages are not unique.
I didn't even realize that `unique()` exists in both convex-helpers and the regular convex library. I saw the two different error messages in my logs and was very confused how `unique()` was ever called from convex-helpers in my repo. I traced it down to me using `customMutation/customQuery` from `convex-helpers/server/customFunctions.ts`.

nipunn1313:
landed with
https://github.com/get-convex/convex-backend/commit/45083425954386d23f73e88ff85dfd3182fca1a0

PR #60, opened 2025-03-14
(docs: nextjs QS) fix incorrect title; update convexclientprovider explanation

alexisintech:
Clerk employee here 👋  While checking out the Convex docs to see how to integrate Convex into a Clerk + Next.js app, I noticed the Next.js QS could use some love. From the Clerk Docs team to yours 💜 

Notes about the changes:
- This all actually started as just [a title change](https://github.com/get-convex/convex-backend/compare/main...alexisintech:convex-backend:main?expand=1#diff-d0f1ae84b677c8e3045bd950b798b2e273145a91bb686efb419883e4f01c16dbR51), I saw "Create a React app" and I was like, well actually you're creating a Next.js app. So I went to open a PR, and was like, well while I'm here, I'll just update a few more things...
- Organize order of events sequentially, for better UX. E.g. "in the `convex/` folder" should be the lead of a sentence, because its the first in a sequence of events. A user will first need to go to their `convex/` folder to then create a new file; it's the first step so it should be the first direction given in the sentence.
- [The colon change in ``and the export name: `api.tasks.get`.``](https://github.com/get-convex/convex-backend/compare/main...alexisintech:convex-backend:main?expand=1#diff-d0f1ae84b677c8e3045bd950b798b2e273145a91bb686efb419883e4f01c16dbR122). It originally read as ``the export name, `api.tasks.get`.`` which grammatically was telling me that the export name was `api.tasks.get` which I was really confused by. When I finally got to the end of the QS, I learned that the API function itself is what's called `api.tasks.get` and then I understood what the doc was trying to convey. So I've made a small change to help understandability.
- Updated the explanation of creating a provider wrapper. I feel the copy was telling me what the code was doing syntactically, but the code speaks for itself. That's why there's an example paired with the copy! So instead, the copy here should be **explaining** what the code is doing. I updated the explanation to the best of my ability, and to how it made sense to me, but if the explanation is incorrect then of course, feel free to update.

Also, I'm surprised you don't abstract the `ConvexClientProvider` part away for users so they can skip the step of creating it themselves and simply import it from Convex itself and then wrap their app (similar to how we do `ClerkProvider`). 👀 

Cheers 😸💖

----

### DIFF
--- a/npm-packages/docs/docs/quickstart/nextjs.mdx
+++ b/npm-packages/docs/docs/quickstart/nextjs.mdx
@@ -48,7 +48,7 @@ Alternatively see the
 version of this quickstart.
 
 <StepByStep>
-  <Step title="Create a React app">
+  <Step title="Create a Next.js app">
     Create a Next.js app using the `npx create-next-app` command.
 
     Choose the default option for every prompt (hit Enter).
@@ -66,8 +66,8 @@ version of this quickstart.
   </Step>
   <Step title="Install the Convex client and server library">
     To get started, install the `convex`
-    package which provides a convenient interface for working
-    with Convex from a React app.
+    package, which provides a convenient interface for working
+    with Convex.
 
     Navigate to your app and install `convex`.
 
@@ -106,9 +106,7 @@ version of this quickstart.
   </Step>
 
   <Step title="Add the sample data to your database">
-    Now that your project is ready, add a `tasks` table
-    with the sample data into your Convex database with
-    the `import` command.
+    Use the [`import`](/database/import-export/import) command to add a `tasks` table with the sample data into your Convex database.
 
     ```
     npx convex import --table tasks sampleData.jsonl
@@ -117,12 +115,11 @@ version of this quickstart.
   </Step>
 
   <Step title="Expose a database query">
-    Add a new file <JSDialectFileName name="tasks.ts" /> in the `convex/` folder
-    with a query function that loads the data.
+    In the `convex/` folder, add a new file <JSDialectFileName name="tasks.ts" /> with a query function that loads the data.
 
     Exporting a query function from this file
     declares an API function named after the file
-    and the export name, `api.tasks.get`.
+    and the export name: `api.tasks.get`.
 
     <TSAndJSSnippet
       sourceTS={tasks}
@@ -133,8 +130,9 @@ version of this quickstart.
   </Step>
 
   <Step title="Create a client component for the Convex provider">
-    Add a new file <JSDialectFileName name="ConvexClientProvider.tsx" /> in the `app/` folder. Include the `"use client";` directive, create a
-    `ConvexReactClient` and a component that wraps its children in a `ConvexProvider`.
+    For `<ConvexProvider>` to work on the client, `<ConvexReactClient>` must be passed to it.
+
+    In the `app/` folder, add a new file <JSDialectFileName name="ConvexClientProvider.tsx" /> with the following code. This creates a component that wraps `<ConvexProvider>` and passes it the `<ConvexReactClient>`.
 
     <TSAndJSSnippet
       sourceTS={ConvexClientProvider}
@@ -145,7 +143,7 @@ version of this quickstart.
   </Step>
 
   <Step title="Wire up the ConvexClientProvider">
-    In <JSDialectFileName name="app/layout.tsx" ext="js" />, wrap the children of the `body` element with the `ConvexClientProvider`.
+    In <JSDialectFileName name="app/layout.tsx" ext="js" />, wrap the children of the `body` element with the `<ConvexClientProvider>`.
 
     <TSAndJSSnippet
       sourceTS={layout}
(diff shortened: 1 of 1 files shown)


npm-packages/docs/docs/quickstart/nextjs.mdx
```diff
@@ -133,8 +130,9 @@ version of this quickstart.
   </Step>
 
   <Step title="Create a client component for the Convex provider">
-    Add a new file <JSDialectFileName name="ConvexClientProvider.tsx" /> in the `app/` folder. Include the `"use client";` directive, create a
-    `ConvexReactClient` and a component that wraps its children in a `ConvexProvider`.
+    For `<ConvexProvider>` to work on the client, `<ConvexReactClient>` must be passed to it.
```
thomasballinger, line 133:
```suggestion
    For `<ConvexProvider>` to work on the client, `ConvexReactClient` must be passed to it.
```

npm-packages/docs/docs/quickstart/nextjs.mdx
```diff
@@ -66,8 +66,8 @@ version of this quickstart.
   </Step>
   <Step title="Install the Convex client and server library">
     To get started, install the `convex`
-    package which provides a convenient interface for working
-    with Convex from a React app.
+    package, which provides a convenient interface for working
+    with Convex.
```
thomasballinger, line 70:
```suggestion
    package.
```
agree "from a React app" wasn't very useful, and now the rest doesn't seem useful either IMO

npm-packages/docs/docs/quickstart/nextjs.mdx
```diff
@@ -133,8 +130,9 @@ version of this quickstart.
   </Step>
 
   <Step title="Create a client component for the Convex provider">
-    Add a new file <JSDialectFileName name="ConvexClientProvider.tsx" /> in the `app/` folder. Include the `"use client";` directive, create a
-    `ConvexReactClient` and a component that wraps its children in a `ConvexProvider`.
+    For `<ConvexProvider>` to work on the client, `<ConvexReactClient>` must be passed to it.
+
+    In the `app/` folder, add a new file <JSDialectFileName name="ConvexClientProvider.tsx" /> with the following code. This creates a component that wraps `<ConvexProvider>` and passes it the `<ConvexReactClient>`.
```
thomasballinger, line 135:
```suggestion
    In the `app/` folder, add a new file <JSDialectFileName name="ConvexClientProvider.tsx" /> with the following code. This creates a client component that wraps `<ConvexProvider>` and passes it the `<ConvexReactClient>`.
```

npm-packages/docs/docs/quickstart/nextjs.mdx
```diff
@@ -172,7 +170,7 @@ version of this quickstart.
   </Step>
 
   <Step title="Start the app">
-    Start the app, open [http://localhost:3000](http://localhost:3000) in a browser,
+    Run your development server, open [http://localhost:3000](http://localhost:3000) in a browser,
```
thomasballinger, line 173: since "development server" is an overloaded term in convex I'll add "Next.js" or "frontend" here

thomasballinger (approved):
Thank you! I'll land this, expect to see it out later today

thomasballinger:
> So instead, the copy here should be explaining what the code is doing

love it

thomasballinger:
> Also, I'm surprised you don't abstract the ConvexClientProvider part away for users so they can skip the step of creating it themselves and simply import it from Convex itself and then wrap their app (similar to how we do ClerkProvider). 

Yeah good point, we held off to avoid needing to create Next.js-specific code using the (at the time only used by) Next.js `"use client"` directives with the goal of keeping the convex package very general; but now that we have some Next.js-specific code in `convex/nextjs` it seems like time to do this! The Clerk nextjs package has been helpful to look at for how to build great libraries that work well with Next.js.

Maybe soon! This isn't the only way to set this up Convex providers with Clerk but it seems reasonable to make the common path easier.

thomasballinger:
Landed in https://github.com/get-convex/convex-backend/commit/85fe152b6923da4ef0bcf62450b65a382bb85a5e

Thanks @alexisintech!

alexisintech:
> Landed in [85fe152](https://github.com/get-convex/convex-backend/commit/85fe152b6923da4ef0bcf62450b65a382bb85a5e)
> 
> Thanks @alexisintech!

woah what did you do to get it into that commit??

also, love all the helpful communication and openness to accepting these updates ❤️  thanks so much :)

thomasballinger:
thanks for sharing the motivation behind these changes, gives us some positive second order effects from teaching in addition to the specific improvements!

we're using [copybara](https://github.com/google/copybara) to get code between our open source repo and the closed-source one, which unfortunately means the PR shows as Closed instead of Merged here but at least does preserve authorship via `Co-authored-by:`

alexisintech:
awesome! i've never seen that, its so cool

PR #86, opened 2025-05-06
docs(tanstack-quickstart): change start status from alpha to beta

prashantbhudwal:
This fixes an inconsistency in the TanStack Start quickstart docs, where the Admonition panel still labeled the project as “Alpha” even though the copy text referred to “Beta.” Since TanStack Start has officially entered beta, this change updates the Admonition title from “Alpha” to “Beta” so that the documentation accurately reflects its current stage.

----


thomasballinger:
@prashantbhudwal thanks! I know Start was in Beta about 5 months ago but given the significant overhaul I wasn't sure it still was; but indeed that's what the website says, so let's go with it!

nipunn1313:
merged in https://github.com/get-convex/convex-backend/commit/85e406d7b6f15f069f3aa667909274ef3aa1418d

PR #92, opened 2025-05-14
Update Svelte docs

didier:
Just some small tweaks that removes reference to "svelte 5 preview" and the old `npm create` CLI in favor of the current `npx sv create` pattern, as well as a wrong reference of `src/routes/page.tsx`.

----

### DIFF
--- a/npm-packages/docs/docs/quickstart/svelte.mdx
+++ b/npm-packages/docs/docs/quickstart/svelte.mdx
@@ -15,18 +15,15 @@ Learn how to query data from Convex in a Svelte app.
 
 <StepByStep>
   <Step title="Create a SvelteKit app">
-    Create a SvelteKit app using the `npx create svelte@latest` command.
+    Create a SvelteKit app using the `npx sv create` command.
 
-    Other sets of options will work with the library as long as Svelte 5 is used but for this quickstart guide:
-
-    - For "Which Svelte app template," choose **"Skeleton project."**
-    - For "Add type checking with TypeScript," choose **"Yes, using TypeScript syntax."**
-    - For "Select additional options," enable **"Try the Svelte 5 preview."**
+    - For "Which template would you like?" choose **"SvelteKit minimal".**
+    - For "Add type checking with TypeScript," choose **"Yes, using TypeScript syntax".**
 
     <br></br>
 
     ```sh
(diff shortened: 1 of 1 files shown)


npm-packages/docs/docs/quickstart/svelte.mdx
```diff
@@ -35,7 +32,7 @@ Learn how to query data from Convex in a Svelte app.
     To get started, install the `convex` and `convex-svelte` packages.
 
     ```sh
-    cd my-app && npm install convex convex-svelte
+    cd my-app && npm install -D convex convex-svelte
```
Nicolapps, line 35: I think that this change shouldn’t be applied because `convex` and `convex-svelte` are used at runtime.

```suggestion
    cd my-app && npm install convex convex-svelte
```

didier: In Svelte everything gets compiled in, it's normal for everything to be a dev dependency unless you specifically want something to be externalized

I have multiple large repos (not Convex) with only devDependencies. 

It's worth testing if this correctly compiles for prod too, but I'm pretty sure it will

nipunn1313: @thomasballinger  ?

thomasballinger: We can try this, but is there any downside to putting it in deps instead of dev deps @didier? This distinction is still made in some svelte apps, in which case these are deps, not dev deps (the convex cli is a dev dep, but it's the same package as convex/server (compiled into convex functions) and convex/browser (compiled into svelte client and (I hope? with every adapter?) bundled into server bundles). If a developer publishes their package to reuse functionality in lib/ then they'll probably need to move convex back to deps.

I thought there were issues where the server-side code is not bundled in some situations, maybe that's all been resolved as server endpoints are always bundled?

I'd rather not just so we can skip the step where we explain that if you reuse code from this package you need to move it.

didier: THere's no downside to putting it in regular deps IIRC, as far as I know Svelte doesn't care that much.

But yes from a docs perspective let's keep it simple!

didier:
Just saw you guys merged the updates elsewhere. Thanks!

Issue #103, opened 2025-05-22
Convex deploy problem on nextjs apps


devLeopar:
Hi;
There seems a bug happening on while deploying convex on either locally or when trying to deploy on vercel.
The documentation suggest using custom build command which is
`npx convex deploy --cmd 'npm run build'`

However, this command fails generating `convex/_generated` folder before nextjs try to build whole app. Therefore, it results in 'module not found' errors on the files that imports modules from this folder.

I found a solution. If we ignore --cmd option and update the custom build command like below:
`npx convex deploy && npm run build`
It waits convex to deploy and generate related folder and files, then run next build.

My suggestion is updating documentation which is here: [https://docs.convex.dev/production/hosting/vercel](https://docs.convex.dev/production/hosting/vercel)

nipunn1313:
Can you provide repro steps? Like what is an example app (ideally one of the Convex demo apps) and add screenshots / exact information from the errors you are seeing.

Thanks for the reprot.

devLeopar:
@nipunn1313 I've created a [PR](https://github.com/get-convex/convex-backend/pull/104) for this and explained a bit more

wkd-kapsule:
The issue is probably the docs saying `npm convex deploy --cmd 'npm run build'` but you can see in the screenshot that it actually is `npm convex deploy --cmd='npm run build'` with `=` after `--cmd`. I guess this is the issue because I'm using `=` for my project and have no issue deploying.

thomasballinger:
Either syntax is allowed, these are both valid parsings. It sounds like the problem is you don't want to commit the `convex/_generated` directory, so you need to generate it each time? I suggest trying `convex codegen && npx convex deploy --cmd 'npm run build'` for this. Generally the _convex/generated directory is committed to version control so this isn't an issue.

G19m99:
I just had this issue while using Netlify and Vite. The issue was that I did not include the `_generated` folder in source control, so there's 2 solutions.

~~1. updated build cmd to `npx convex deploy && npm run build` and also had to add the `VITE_CONVEX_URL` (or in your case, the nextjs version of that) to Netlify~~ see @ianmacartney comment
2. update cmd to `npx convex codegen && npx convex deploy --cmd 'npm run build'`

Both of these should give the app time to generate the code needed before running the build cmd (i personally went with option 1 and it worked).

[docs](https://docs.convex.dev/cli#update-generated-code) says here it should be included in source control

ianmacartney:
[...]

PR #104, opened 2025-05-22
fix: replacing --cmd with && in custom build commands

devLeopar:
This should resolve https://github.com/get-convex/convex-backend/issues/103

here is some more information about this
image

This _generated folder never be created with the --cmd option. This is not a file path issue. I've tried combinations on next build however, the issue is the folder absence. It is replicated locally as well. You can use your `CONVEX_DEPLOY_KEY` and run that command to see

----

### DIFF
--- a/npm-packages/docs/docs/production/hosting/vercel.mdx
+++ b/npm-packages/docs/docs/production/hosting/vercel.mdx
@@ -30,7 +30,7 @@ follow the [Convex React Quickstart](/quickstart/react.mdx) first. Then:
   </Step>
   <Step title="Override the Build command">
     Override the "Build command" to be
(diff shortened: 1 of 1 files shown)


thomasballinger:
It sounds like you've got a custom setup where you don't commit the `_generated` file, generally it's recommended that already exist in version control.

The point of this --cmd is to avoid deploying the convex functions if the frontend build fails:
1. set .env.local environment variable
2. run frontend build
3. if that succeeded, build and deploy convex code.

We could change this a bit for setups like yours that dont' commit _generated by inserting a step between 1 and 2 that runs codegen, but the suggestion in the PR to unconditionally deploy convex functions isn't workable, we don't want to end up deploying new convex functions without deploying a new frontend because the frontend build is broken.

PR #108, opened 2025-05-28
print tablename in unique() error message

djbalin:
Improves usefulness of the error thrown when `.unique()` returns more than 1 result by now also printing the table name.

Improves quality of life: currently, it can be cumbersome to identify the actual table in which these non-unique elements belong.

This solution is a bit work-around-y, but it's the only solution I could come up with. Let me know if there is a better way to achieve this!

----

### DIFF
--- a/npm-packages/convex/src/server/impl/query_impl.ts
+++ b/npm-packages/convex/src/server/impl/query_impl.ts
@@ -166,9 +166,15 @@ export class QueryImpl implements Query<GenericTableInfo> {
     | { type: "executing"; queryId: number }
     | { type: "closed" }
     | { type: "consumed" };
-
+  private tableName: string;
+  
   constructor(query: SerializedQuery) {
     this.state = { type: "preparing", query };
+    if (query.source.type === "FullTableScan") {
+      this.tableName = query.source.tableName;
+    } else {
+      this.tableName = query.source.indexName.split(".")[0];
+    }
   }
 
   private takeQuery(): SerializedQuery {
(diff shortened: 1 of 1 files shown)


npm-packages/convex/src/server/impl/query_impl.ts
```diff
@@ -333,7 +339,7 @@ export class QueryImpl implements Query<GenericTableInfo> {
       return null;
     }
     if (first_two_array.length === 2) {
-      throw new Error(`unique() query returned more than one result: 
+      throw new Error(`unique() query returned more than one result from table ${this.tableName}:
```
nipunn1313, line 342: @thomasballinger what do you think?
I almost feel like it would be more useful to say "unique() returned more than one result from query on ${this.indexName}"

nipunn1313:
thx for the idea! This looks reasonable - just want to check in with tom.

thomasballinger:
Side note, if there's anywhere that developers see this error without a stack trace that's something we should fix, this is a good improvements to make as well but making sure there's a stack trace is top pri

npm-packages/convex/src/server/impl/query_impl.ts
nipunn1313: @djbalin what do you think about printing indexName here instead of tableName? It'll give more information.

djbalin:
> Side note, if there's anywhere that developers see this error without a stack trace that's something we should fix, this is a good improvements to make as well but making sure there's a stack trace is top pri

@thomasballinger I haven't experienced missing stack trace, pasted is a log example. As also outlined in my other comment, I suppose that a developer who runs into `.unique()` throwing this error will want to investigate those unexpectedly non-unique data points in their Convex dashboard. Finding the relevant piece of code through the stack trace takes a few more steps, and implementation may have changed or move in the meantime. I think having the error message point directly to the actual raw implementation-agnostic data would be useful

image

thomasballinger:
Let's do it, thanks @djbalin! This is great, I think we have other error messages that coudl be better like this.

djbalin:
Awesome :) there is also an implementation of `.unique()` in `convex-helpers`, would be nice to update that as well with whatever you end on!

thomasballinger:
This is in https://github.com/get-convex/convex-backend/commit/d8247687d8108e1bc11a228eae5f51c18144fdca, thanks @djbalin!

PR #113, opened 2025-06-01
Fix typo in`zen.mdx` - `/understanding/zen`

galer7:
----


victorpatru:
There is an expression "Don't go at it alone" btw

thomasballinger:
Yeah this is written as intended @galer7. You mention conciseness but the change here isn't any more concise. If you're finding docs confusing we'd love to hear, let us know about the confusion and we'll figure something out.

PR #124, opened 2025-06-20
docs: remove redundant line in mutation docs

efekrskl:
Removes repeated sentence from mutation docs to keep it consistent with query docs.

Thanks for the great tool and docs!

----


nipunn1313:
landed in https://github.com/get-convex/convex-backend/commit/32e33a40b8747184a9133db9ef558a9fb2a0f008

Issue #126, opened 2025-06-24
Support convex.config.[cm]js and convex.config.d.[cm]ts


prisis:
Currently convex components support only the .js and .d.ts extensions.

To support standard extensions for esm and cjs, convex should check if a convex.config does exists with .cts,.mts and .d.cts,.d.mts.

thomasballinger:
@prisis is there a reason you'd like these to be supported? convex.config.js must be esm format so there's no need to support .cjs, and components always use "type": "module" so the extension shouldn't be necessary.

If you're authoring components that's great, the interface is subject to change but we do want this to be accessible. See https://github.com/get-convex/templates/tree/main/template-component for the current recommended template. If there's something blocking you here would love to hear about it.

prisis:
I've been working on a fork of the @convex-dev/agent package to extend its functionality.

My main goal was to move away from the complex configurations often required by tsc to handle both CommonJS and ESM outputs. The standard approach often leads to issues like:

- Including unnecessary files in the final package (e.g., vitest.js in the dist folder https://www.npmjs.com/package/@convex-dev/agent?activeTab=code).
- tsc struggling to correctly generate type definition files (.d.ts) for files that have both default and named exports.
- Needing extra package.json files in subdirectories just to declare the module type ("type": "module"), which adds clutter.
- There are a lot of edge cases, that tsc cant handle, one more would be the typesVersions key, generation.

To address this, I used a modern bundler, [packem](https://github.com/visulima/packem/tree/main/packages/packem), which simplifies this process immensely. By using explicit file extensions (.mjs for ESM, .cjs for CommonJS) there are other like rolldown, tsup etc, the module type is clear, and the need for extra package.json files is eliminated.

This refactor, which you can see in [my fork here](https://github.com/anolilab/anole.chat/tree/main/packages/ai-agent/package.json), successfully reduced the package size from 2.93 MB to 2.2 MB (with esm and cjs support) with esm only support its 1.6 MB.

However, I've run into a blocker: Convex doesn't seem to recognize .mjs and .d.mts files. This prevents me from using this more modern, efficient package structure within the Convex environment.

thomasballinger:
Ah thanks for the context, @ianmacartney mentioned this the other day, cool that you're looking at this.

More important than the package size, using more standard tools here would be great.

Ideally we could get some test coverage if we're changing things, https://github.com/get-convex/convex-backend/tree/main/npm-packages/private-demos/components-poc has some of this.

[...]

PR #127, opened 2025-06-24
feat: enhance file resolution logic for Convex 

prisis:
configuration and schema files

- Updated bundler to check for multiple file extensions (.ts, .js, .mjs) for schema files.
- Improved component root path resolution to prioritize new file types.
- Adjusted comments and error messages for clarity regarding supported file types.

This change ensures better compatibility and flexibility in handling various configuration file formats.

fixes https://github.com/get-convex/convex-backend/issues/126

----


npm-packages/convex/src/bundler/index.ts
```diff
@@ -297,10 +297,26 @@ export async function bundleSchema(
   dir: string,
   extraConditions: string[],
 ) {
-  let target = path.resolve(dir, "schema.ts");
-  if (!ctx.fs.exists(target)) {
-    target = path.resolve(dir, "schema.js");
+  const candidates = [
```
thomasballinger, line 300: would love to hear if people are using these extensions and why, but if this is helpful we can make this change. I'd guess the various TypeScript extensions would be more important? .mts, .cts, etc than .js, since most users use TypeScript? But I haven't heard requests for either

npm-packages/convex/src/bundler/index.ts
```diff
@@ -314,21 +330,30 @@ export async function bundleSchema(
...
+  const existingPaths = candidates
+    .map(filename => path.resolve(dir, filename))
+    .filter(filepath => ctx.fs.exists(filepath));
+  
+  if (existingPaths.length > 1) {
     return await ctx.crash({
       exitCode: 1,
       errorType: "invalid filesystem data",
-      printedMessage: `Found both ${authConfigPath} and ${authConfigTsPath}, choose one.`,
+      printedMessage: `Found multiple auth config files: ${existingPaths.join(", ")}, choose one.`,
```
thomasballinger, line 348: this could regress some (probably broken) workflows where people compile their ts to js in-place

npm-packages/convex/src/cli/lib/components.ts
```diff
@@ -54,19 +54,27 @@ import { Reporter, Span } from "./tracing.js";
 import {
   DEFINITION_FILENAME_JS,
   DEFINITION_FILENAME_TS,
+  DEFINITION_FILENAME_MJS,
+  DEFINITION_FILENAME_CJS,
```
thomasballinger, line 58: the component changes I do not want to make without understanding the use cases. The spec for components isn't well-defined yet but I wouldn't want to make what's accepted anyy more broad than necessary without specifics use cases

thomasballinger (commented):
I don't think we want the component-related changes, but if anyone has a reason to need to use alternative file extensions for files like schema and auth.config we can do that; just want to hear about the uses cases first so we can add realistic tests of them and we know what we're supporting

npm-packages/convex/src/bundler/index.ts
thomasballinger: This is fine if we remove .cjs, we never want to bundle a commonjs schema.

npm-packages/convex/src/bundler/index.ts
```diff
@@ -314,21 +330,30 @@ export async function bundleSchema(
 }
 
 export async function bundleAuthConfig(ctx: Context, dir: string) {
-  const authConfigPath = path.resolve(dir, "auth.config.js");
-  const authConfigTsPath = path.resolve(dir, "auth.config.ts");
-  if (ctx.fs.exists(authConfigPath) && ctx.fs.exists(authConfigTsPath)) {
+  const candidates = [
```
thomasballinger, line 333: auth.config.ts doesn't exist in components so there's no need to change this

npm-packages/convex/src/cli/lib/components.ts
thomasballinger: We do not want the CJS one. I don't know that we really need a .mjs extension either, but if this is a difficult limitation for a packaging tool we can add .mjs if we also add a component that works like this to make sure we don't regress the behavior. 

If you'd like to use packem to build convex components, would you consider adding the ability to produce .js files for ESM-only builds of a package with `"type": "module"`? e.g. with `tsdown`, an ESM-only build configured like below produces ESM files with .js extensions. Component definitions are already always ESM, so it's never necessary to distinguish between these.

```
import { defineConfig } from "tsdown";

export default [
  defineConfig({
    entry: ["src/index.ts", "src/another.ts"],
    format: ["cjs", "esm"],
    dts: {
      tsconfig: "./tsconfig.json",
      sourcemap: true,
    },
  }),
  defineConfig({
    entry: ["src/server.ts"],
    format: ["esm"],
    dts: {
      tsconfig: "./tsconfig.json",
      sourcemap: true,
    },
  }),
];
```

npm-packages/convex/src/cli/lib/components.ts
```diff
@@ -54,19 +54,27 @@ import { Reporter, Span } from "./tracing.js";
...
+  for (const filename of candidates) {
+    const componentRootPath = path.resolve(path.join(functionsDir, filename));
+    if (ctx.fs.exists(componentRootPath)) {
+      return componentRootPath;
+    }
   }
-  return componentRootPath;
+  
+  // Default fallback to .ts for backward compatibility
+  return path.resolve(path.join(functionsDir, DEFINITION_FILENAME_TS));
```
thomasballinger, line 77: this is a behavior change, previously the default was the JS path

npm-packages/convex/src/cli/lib/components/definition/bundle.ts
```diff
@@ -53,7 +53,7 @@ function componentPlugin({
     name: `convex-${mode === "discover" ? "discover-components" : "bundle-components"}`,
     async setup(build) {
       // This regex can't be really precise since developers could import
-      // "convex.config", "convex.config.js", "convex.config.ts", etc.
+      // "convex.config", "convex.config.mjs", "convex.config.cjs", "convex.config.js", "convex.config.ts", etc.
```
thomasballinger, line 56: same as above, throughout the PR remove .cjs

prisis:
So i did remove the support for the .cjs and .d.cts, and reverted the order of finding back to "ts", "mjs" "js".  :)

It still could be that a .js is a commonjs file. But there is not easy way to check if a file is esm only (i dont know one)

prisis:
Hey @thomasballinger ,

Wanted to know what are the plans for this pr?

Issue #130, opened 2025-06-27
Remote IP address in queries/mutations/actions


SuperJakov:
Some compliance and certification processes (such as audit logging or user tracing) require logging the remote IP address of the client making the request. 

Currently, there doesn’t appear to be a built-in way to retrieve the client's IP address using the `ctx` object in Convex actions or functions.

**Request:**  
It would be helpful if Convex provided a reliable and secure way to access the remote IP address of a request via `ctx`, so developers can log or use it as needed for auditing or security-related purposes.

**Use case example:**  
```ts
export const someAction = action({
  args: { ... },
  handler: async (ctx, args) => {
    const ip = ctx.request?.ip; // Ideally something like this
    logToAuditTrail({ userId: ctx.user?.id, ip });
  },
});
```

Thank you for considering this!

thomasballinger:
Agree this should be in. For httpActions, no problem: we already know the way to expose them, additional headers, something like `X-Forwarded-For`. Just prioritization, this coming.

For Actions, sometimes an action is scheduled doesn't have an IP associated with it; we could treat this similar to auth where it's your job to propagate it to a scheduled functions if you need it there. A little more design required, could be `ctx.request` or `ctx.client` or `ctx.meta`, I like `ctx.request`. Depending on the transport it might be a WebSocket, it might be an HTTP request, it might be scheduled; so you can't count on it being there, but yeah we can probably do this.

For mutations, same deal but it gets a bit tricky because mutations can retry, need to figure out which IP to use. Could always be the websocket connection in which the mutations was initially enqueued, not too bad? 

For queries, this is a great way to make your query very unlikely to get cache hits. Maybe, but we're not starting with this. We'd need to track whether you accessed it. Doable stuff, doesnt' exist yet, not coming soon.

nipunn1313:
See #132 for a community implementation. I think we probably want to hash out the design (per what @thomasballinger laid out) before committing to a solution here.

If I'm being honest, we probably won't have time to prioritize this right away - but it's definitely on our list.

typed-sigterm:
any progress?

ianmacartney:
We're approaching this from two different angles:
- Audit logs will be able to opt-in to putting IP/userAgent in the log automatically
[...]

PR #132, opened 2025-06-28
Remote IP available through context-object

COCPORN:
## PURPOSE

- Enrich the context-object with `getRemoteIp()` for `query`/`mutation`/`action`, returning the remote IP on the connection
- "Zero cost" if not used, called synchronously for speed
- Opens for application level audit-logging, IP-based feature-blocking/rate-limiting, etc

## CAVEAT

Depending on how Convex Cloud is load-balanced, this might or might not work. If the service is setup using Layer 7 load balancers, this will not work as it relies on `X-Forwarded-For` being set, which this PR doesn't honor (but can be made to, conceptually).

If it is balanced using Layer 4 with NLB, this will _probably work with Convex Cloud_ as it usually passes on the source IP-address directly.

_It will very likely work_ for standalone backend.

**Also:** I am not sure it needs to be implemented as a function call, but it was not obvious how to have this available when V8 is initialized. Implementing it as a property directly on `ctx` would incur some unwanted overhead as JavaScript needs to call back into Rust, which we want to avoid in the common case.

## USAGE

```typescript
  // In your Convex functions:
  export const logClientInfo = mutation({
    handler: async (ctx, args) => {
      // Only pays cost when you actually need it
      const clientIp = ctx.getRemoteIp();
      console.log("Request from: ", clientIp);

      // Functions that don't call getRemoteIp() have "zero overhead"
      return { success: true };
    }
  });
```

## TEST RUN

```
test result: FAILED. 154 passed; 3 failed; 5 ignored; 0 measured; 0 filtered out; finished in 10.62s
```

For me, this is currently _the same_ as the `main`-branch, so I don't think errors are introduced that tests would catch.

## CODE QUALITY

This PR needs to be looked after by a seasoned Convex developer as it is my first attempt at making something in Convex core. If it is rejected, feel free to use it as inspiration for a proper implementation.

## IMPROVEMENTS (potential)

- Split into `getRemoteIp` and `getRemotePort`, perhaps, currently the port is returned with `getRemoteIp` on the format `127.0.0.1:57345`.
----

### DIFF
--- a/crates/application/src/cron_jobs/mod.rs
+++ b/crates/application/src/cron_jobs/mod.rs
@@ -639,6 +639,7 @@ impl<RT: Runtime> CronJobContext<RT> {
                     request_id.clone().unwrap_or_else(RequestId::new),
                     execution_id.clone().unwrap_or_else(ExecutionId::new),
                     caller.parent_scheduled_job(),
+                    caller.remote_ip(),
                     caller.is_root(),
                 );
                 sentry::configure_scope(|scope| context.add_sentry_tags(scope));
--- a/crates/application/src/scheduled_jobs/mod.rs
+++ b/crates/application/src/scheduled_jobs/mod.rs
@@ -832,6 +832,7 @@ impl<RT: Runtime> ScheduledJobContext<RT> {
                     request_id.clone().unwrap_or_else(RequestId::new),
                     execution_id.clone().unwrap_or_else(ExecutionId::new),
                     caller.parent_scheduled_job(),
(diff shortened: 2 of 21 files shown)


COCPORN:
BTW, I seem to have introduced a compile time warning, I will look at that today.

nipunn1313:
thx for the PR. Check out the discussion in #130 - too.

v0id-user:
Hey, just wondering if there are any updates on this... looks super useful :)

Issue #140, opened 2025-07-07
[Bug] Convex CLI fails on windows due to the `:` in path


arduano:
This line gets hit when trying to run `convex dev` on windows: https://github.com/get-convex/convex-backend/blob/5536b756d65774b7185de05e1bb6f903068075b0/crates/convex/sync_types/src/path.rs#L25

Here's a sample (simplified) error that I get when I run it:
```
✖ Error fetching POST  http://my.convex.instance:3210/api/prepare_schema 400 Bad Request: BadConvexModuleIdentifier: C:/path/to/repo/out/schema.js is not a valid path to a Convex module. Path component C: can only contain alphanumeric characters, underscores, or periods.
```

I'm not sure what the ideal solution would be, but I'm confused why a server-side error would be happening due to dev-side configuration. Maybe using relative paths relative to the closes .git folder would be better?

nipunn1313:
Agreed! Likely a bug somewhere. The idea is that a "Convex Module Identifier" should be something platform agnostic.

Clientside path (samples)
C:\nipunn\repo\convex\dir\file.ts
/Users/nipunn/repo/convex/dir/file.ts

Module path
dir/file.js

The CLI should be converting these properly before it gets to the server. There's probably a bug somewhere. Do you have a repro? Like a small repo that hits this issue?

Alternately, do you have the name of the cloud deployment that you're developing against? If so, then we can look at logs there. Should be in your `.env.local`.

arduano:
I'm deploying against self-hosted because I'm just testing the feasibility of some things and learning more about Convex. Unfortunately I'm stuck with windows in my enterprise-scale company for more reasons than one haha

I'm using these images: `ghcr.io/get-convex/convex-backend:6efab6f2b6c182b90255774d747328cfc7b80dd9` `ghcr.io/get-convex/convex-dashboard:6efab6f2b6c182b90255774d747328cfc7b80dd9`

My repro would just be: spin up a local node repo with convex on windows, run `convex dev`

arduano:
I abbreviated to `my.convex.instance:3210` in my error because it's just a local IP address

nipunn1313:
self-hosting on windows is not as battle tested as on other platforms (since we run cloud convex on linux ourselves).

Windows should work, but I'm not surprised to find little bugs like this. We'd love to get self-hosted windows to a stronger state, but in the meantime you can probably try testing things out with cloud convex dev deployment.

[...]

PR #148, opened 2025-07-11
fix(cli): normalize Windows paths for module identifiers and add test…

kmr-rohit:
… (#140)

fix(cli): issue #140 - normalize Windows paths for module identifiers and added tests for windows paths. 

----

### DIFF
--- a/npm-packages/convex/src/bundler/index.ts
+++ b/npm-packages/convex/src/bundler/index.ts
@@ -206,8 +206,7 @@ export async function bundle(
       sourceMaps.set(relPath, outputFile.text);
       continue;
     }
-    const posixRelPath = relPath.split(path.sep).join(path.posix.sep);
-    const normalizedPath = normalizeModulePath(posixRelPath);
+    const normalizedPath = normalizeModulePath(outputFile.path);
     modules.push({ path: normalizedPath, source: outputFile.text, environment });
   }
   for (const module of modules) {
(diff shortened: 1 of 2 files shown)


kmr-rohit:
@nipunn1313  , can you look into it.

nipunn1313:
Been a little busy - will try to get to it with some time. Thanks for the contribution.

npm-packages/convex/src/bundler/index.ts
```diff
@@ -437,6 +438,13 @@ export async function entryPoints(
 // A fallback regex in case we fail to parse the AST.
 export const useNodeDirectiveRegex = /^\s*("|')use node("|');?\s*$/;
 
+export function normalizeModulePath(posixRelPath: string): string {
+  // Remove Windows drive letter (e.g., "C:/" or "D:/") if present
+  let noDrive = posixRelPath.replace(/^[A-Za-z]:\\|^[A-Za-z]:\//, "");
+  // Convert all backslashes to forward slashes
+  return noDrive.replace(/\\/g, "/");
+}
```
nipunn1313, line 446: I suspect we're going to want to use standard libraries rather than doing our own custom string crunching like this.

Windows paths are incredibly complex and there are a lot of ways they can be represented - sticking to standard libraries rather than writing custom will be worthwhile.

https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats


Would also appreciate if you could look into where the module path is used and why it's sent up to the server. That may help us understand what the right solution is. In general, having OS-specific paths sent to the server is probably not the right architecture since the same convex backend can be developed with from multiple OSes.

kmr-rohit: Thanks @nipunn1313  , i will take a look at standard libraries for that, also will check usage & why it is sent up to server. Thanks for the input

kmr-rohit:
Hey @nipunn1313, thanks for the feedback! You're absolutely right about using standard libraries and understanding the architecture better.

I dug into how module paths actually work and found some interesting things:

**The real issue**: Module paths aren't file paths at all - they're logical identifiers like `"convex/myFunction.js"` that the server uses internally. The CLI is supposed to convert OS file paths into these platform-agnostic identifiers before sending them over.

Looking at the server code in `module_path.rs`, it's pretty strict about what it accepts:
- Must be relative paths (rejects anything absolute)
- Forward slashes only  
- No `..` components
- Has to pass `check_valid_path_component()` validation

So when someone on Windows has `C:/path/to/out/convex/foo.js`, we need to extract just the `convex/foo.js` part as the module identifier.

**New approach using standard lib**:
```ts
export function normalizeModulePath(outputPath: string, baseDir: string = "out"): string {
  // Normalize separators first
  const normalizedOutput = outputPath.replace(/\\/g, '/');
  const normalizedBase = baseDir.replace(/\\/g, '/');
  
  const relativePath = path.posix.relative(normalizedBase, normalizedOutput);
  
  // For absolute paths, find where the base dir appears and extract what comes after
  if (relativePath.startsWith('../')) {
    const outputParts = normalizedOutput.split('/');
    const baseParts = normalizedBase.split('/');
    
    const baseIndex = outputParts.findIndex((part, index) => {
      return baseParts.every((basePart, bIndex) => 
        outputParts[index + bIndex] === basePart
      );
    });
    
    if (baseIndex >= 0) {
      return outputParts.slice(baseIndex + baseParts.length).join('/');
    }
    
    return outputParts[outputParts.length - 1];
  }
  
  return relativePath;
}
```

This way we're using `path.posix.relative()` and proper path parsing instead of regex, and it correctly handles the Windows absolute path case by extracting the actual module identifier the server expects.

Tests are all passing now - Windows paths like `C:\project\out\convex\foo.js` properly become `convex/foo.js`.

kmr-rohit:
@nipunn1313  , Had a chance to look into it ?

nipunn1313:
Hi @kmr-rohit - not yet. Lot going on over at Convex - it's in my queue.

kmr-rohit:
Totally understandable, @nipunn1313 — with everything going on at Convex and being early adopters of PlanetScale’s Postgres, it’s all shaping up as it’s meant to be.

ExpDev07:
This is still an issue btw...

drew-wallace:
@nipunn1313 Can you take a look? Still affecting folks.

santiago-ramos-02:
Please, node actions are unusable while developing

PR #155, opened 2025-07-18
Fix: JSON import UTF-8 BOM handling issue

### LINKED ISSUE #138: [Bug] Import | JSON import fails - `Not valid JSON: expected value at line 1 column 1` - due to UTF8-BOM character
Getting `Not valid JSON: expected value at line 1 column 1`  while trying to import as json `npx convex import --table accounts accounts.json`

JSONs formats that I tried:

```json
[
	{
		"accountNumber": "0001",
		"accountName": "reserviert"
	},
	{
		"accountNumber": "0040",
		"accountName": "reserviert"
	}
]
```

```json
{
	"accounts": [
		{
			"accountNumber": "0001",
			"accountName": "reserviert"
		},
		{
			"accountNumber": "0040",
			"accountName": "reserviert"
		}
	]
}

```

Shyam-Raghuwanshi:
Fix JSON import UTF-8 BOM handling issue

- Add strip_utf8_bom() function to remove UTF-8 BOM (EF BB BF) from file content
- Apply BOM stripping to JsonLines and JsonArray import formats
- Add comprehensive tests for UTF-8 BOM handling across all import formats
- Ensure backward compatibility with files without BOM
- Fix 'Not valid JSON: expected value at line 1 column 1' error for BOM files

fix #138 

----

### DIFF
--- a/crates/application/src/snapshot_import/parse.rs
+++ b/crates/application/src/snapshot_import/parse.rs
@@ -183,26 +183,31 @@ pub async fn parse_objects<'a, Fut>(
         ImportFormat::JsonLines(table_name) => {
             let reader = stream_body().await?;
             yield ImportUnit::NewTable(component_path, table_name);
-            let mut buf = Vec::new();
-            let mut reader = reader.into_reader();
-            reader.read_to_end(&mut buf).await?;
-
-            // Check for UTF-8 BOM and reject it
-            if buf.len() >= 3 && buf[0] == 0xEF && buf[1] == 0xBB && buf[2] == 0xBF {
-                anyhow::bail!(ImportError::Utf8BomNotSupported);
-            }
-            let content = std::str::from_utf8(&buf).map_err(|e| {
-                ImportError::NotUtf8(std::io::Error::new(std::io::ErrorKind::InvalidData, e))
-            })?;
-
-            for (lineno, line) in content.lines().enumerate() {
-                let lineno = lineno + 1;
+            let mut reader = tokio::io::BufReader::new(reader.into_tokio_reader());
+            let mut line = String::new();
+            let mut lineno = 1;
+            // Check for UTF-8 BOM at the start of the first line
+            let mut first_line_checked = false;
+            while reader.read_line(&mut line).await.map_err(|e| {
+                ImportError::NotUtf8(e)
+            })? > 0 {
+                // Check for BOM only on the first line
+                if !first_line_checked {
+                    first_line_checked = true;
+                    if line.len() >= 3 && line.as_bytes()[0] == 0xEF && line.as_bytes()[1] == 0xBB && line.as_bytes()[2] == 0xBF {
+                        anyhow::bail!(ImportError::Utf8BomNotSupported);
+                    }
+                }
                 if line.trim().is_empty() {
+                    line.clear();
+                    lineno += 1;
                     continue;
                 }
-                let v: serde_json::Value = serde_json::from_str(line)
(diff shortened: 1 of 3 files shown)


nipunn1313:
Hi.
We ended up fixing clientside with

https://github.com/get-convex/convex-backend/commit/b671cd8ce0d5da17913744d6077f2a474fa4b605

We thought about doing something serverside, but ended up deciding that BOM was not supported by the server - and we preferred solutions that errors if a BOM is sent up.

Currently, it still says
'Not valid JSON: expected value at line 1 column 1'

If you could modify the PR to give a more specific error in this case, we'd take that instead, but we'd like to keep BOM as unsupported.

nipunn1313:
Sorry - I didn't follow up on #138 promptly

crates/application/src/snapshot_import/parse.rs
```diff
@@ -182,16 +181,28 @@ pub async fn parse_objects<'a, Fut>(
...
+                    first_line_checked = true;
+                    if line.len() >= 3 && line.as_bytes()[0] == 0xEF && line.as_bytes()[1] == 0xBB && line.as_bytes()[2] == 0xBF {
+                        anyhow::bail!(ImportError::Utf8BomNotSupported);
+                    }
+                }
+                if line.trim().is_empty() {
+                    line.clear();
+                    lineno += 1;
+                    continue;
+                }
```
nipunn1313, line 205: This seems like an unrelated, undesirable change?

nipunn1313 (changes requested):
overall looks good - thanks
One comment.

crates/application/src/snapshot_import/parse.rs
nipunn1313: I think it's ok for empty lines to be considered bad input (it seems like bad jsonl anyway).

But stepping back - I think it's good practice to have separate changes in separate PRs, so we can have the discussion separately. I'd be happy to approve+merge this w/o further discussion without the blank line change.

If you like the blank line stuff, we can chat about it in a separate PR.

Appreciate the contribution!!

Shyam-Raghuwanshi: Got it. I would like to discuss the `if` logic in a separate PR

Thanks for your time.

crates/application/src/snapshot_import/parse.rs
```diff
@@ -182,16 +181,23 @@ pub async fn parse_objects<'a, Fut>(
             }
         },
         ImportFormat::JsonLines(table_name) => {
-            let mut reader = stream_body().await?.into_reader();
+            let reader = stream_body().await?;
             yield ImportUnit::NewTable(component_path, table_name);
+            let mut reader = tokio::io::BufReader::new(reader.into_tokio_reader());
```
nipunn1313, line 186: curious why this changed. Am just a bit wary of unrelated changes.

I might change this back to the `.into_reader()` strategy from earlier when merging - just to avoid accidental unrelated regressions.

Shyam-Raghuwanshi: You're right. I unnecessarily changed the JsonLines reader setup. This change was outside the scope of the UTF-8 BOM fix. Maybe I made these changes in the beginning

nipunn1313:
merged in https://github.com/get-convex/convex-backend/commit/42a875288fcd598eba4f9a1c7705889b02d710ae

@Shyam-Raghuwanshi - I made some small changes to your PR to simplify. Take a look - trying to really make the diff more minimal from the previous behavior.

PR #164, opened 2025-07-29
Fix patch when exactOptionalPropertyTypes is enabled

sparebytes:
A false Typescript error is generated when setting an optional field to `undefined` if `exactOptionalPropertyTypes` is enabled in the tsconfig.

This PR fixes that by allowing `undefined` to be passed to optional fields when doing a patch.

Example:

```jsonc
// tsconfig.json
{
  "compilerOptions": {
    "exactOptionalPropertyTypes": true,
    "strict": true
  }
}
```

```ts
// schema.ts
export default defineSchema({
  things: defineTable({
    name: v.string(),
    other: v.optional(v.string()),
  }),
});

// things.ts
import { v } from "convex/values";
import { internalMutation } from "./_generated/server";
export const thingMutation = internalMutation({
  args: { id: v.id("things") },
  handler: async (ctx, args) => {
    await ctx.db.patch(args.id, {
      name: "abc",
      other: undefined,
    });
    // Error: Argument of type '...' is not assignable to parameter of type 'Partial<...>' with 'exactOptionalPropertyTypes: true'
  },
});
```

----

### DIFF
--- a/npm-packages/convex/src/server/database.ts
+++ b/npm-packages/convex/src/server/database.ts
@@ -191,7 +191,7 @@ export interface GenericDatabaseWriter<DataModel extends GenericDataModel>
   patch<TableName extends TableNamesInDataModel<DataModel>>(
     table: NonUnion<TableName>,
     id: GenericId<TableName>,
-    value: Partial<DocumentByName<DataModel, TableName>>,
+    value: PatchValue<DocumentByName<DataModel, TableName>>,
   ): Promise<void>;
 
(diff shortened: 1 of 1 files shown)


thomasballinger:
This is in https://github.com/get-convex/convex-backend/commit/07dd1aceaf0b6ff8679fbc162d2ffc5473c7389f, thanks for fixing. No schedule for a new client release atm, likely sometime next week.

PR #182, opened 2025-08-11
Update using-cursor.mdx

rialco:
Hey!

Noticed the documentation for adding Convex docs to cursor is a bit outdated and do not properly reflect the current state of how it should be done.

1. The way to add to docs is now under a different menu `Cursor settings` > `Indexing & Docs` > `Docs`

image

2. It seems that you guys redirect `https://docs.convex.dev/` to `https://docs.convex.dev/home` and therefore when adding the url currently showcased in the docs, cursor its not able to index the documentation. The proposed url makes cursor able to correctly index the 278 pages of docs.

----

### DIFF
--- a/npm-packages/docs/docs/ai/using-cursor.mdx
+++ b/npm-packages/docs/docs/ai/using-cursor.mdx
@@ -106,8 +106,8 @@ projects.
 Adding Convex docs can let you specifically refer to Convex features when
 building your app.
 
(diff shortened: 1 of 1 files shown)


nipunn1313:
yep - looks like cursor changed it. Appreciate the contrib

nipunn1313:
merged in https://github.com/get-convex/convex-backend/commit/6ec3be562eabb010c3da8de36a2a965996478271

PR #183, opened 2025-08-11
Docs: Update the code snippet to be in alignment with the screenshot and the example at large

johncmunson:
docs-fix

### DIFF
--- a/npm-packages/docs/docs/understanding/index.mdx
+++ b/npm-packages/docs/docs/understanding/index.mdx
@@ -203,7 +203,7 @@ WebSocket connection (4):
 
 ```json
 [
-  { _id: "e4g", title: "Grocery shopping", complete: false },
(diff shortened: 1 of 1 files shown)


thomasballinger (approved):

nipunn1313:
hrm wait - I actually think the query in the doc is listing all the incomplete tasks - so it makes sense that it goes from returning 2 items to 1 item. I *think* the original code is correct.

nipunn1313:
Can you double check me here @johncmunson - just want to make sure we do this right.

johncmunson:
Actually... @nipunn1313 is completely right. I had in inverted understanding of the mental model here. In this example, it's not the result of the mutation that will be getting sent down the wire... it's the result of the _dependent query_.

The original example was correct. My bad. Thanks guys.

PR #186, opened 2025-08-14
Update QuickstartsList.tsx

uxbathe:
Add Expo in label


thomasballinger:
Expo isn't required for this this client library, but the quickstart does use it — want to change the Quickstart title instead?

PR #187, opened 2025-08-15
Refactor S3 credential handling in aws_utils

### LINKED ISSUE #185: Use fallback to AWS credential provider
Thanks for a great product!

Self-hosted instances may be in an environment where AWS access keys are discouraged.  Could you remove the requirement to have `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` explicitly set?

If the Rust S3 client uses the [default retrieval order](https://docs.aws.amazon.com/sdk-for-rust/latest/dg/credproviders.html), then the environment variables will still be used first; however, other methods like attaching an IAM policy to an ECS task or EC2 instance will also work (options 4 and 5 in the credential retrieval order).  

I believe the change could be made in the convex-backend/crates/aws_utils/src/lib.rs file to update `must_config_from_env()`.  Maybe you could add another environment variable like `AWS_CREDENTIALS_METHOD` that could override the Convex default and allow the S3 Rust client default.

HeathHopkins:
Refactor S3 credential handling in aws_utils to allow S3 credentials from sources other than environment variables by using the AWS default credential chain.

Preserved credential checking before first use to prevent logs from filling up.

Also added `aws-credential-types` and `tokio` to dependencies in Cargo.toml and aws_utils/Cargo.toml

This PR preserves the same S3 client environment variable credential checks while allowing credential sources from IAM roles which is useful for local debugging (`aws sso`) and self-hosting on AWS (EC2 IAM role attachment).

The AWS S3 Rust SDK uses this default credential chain:  `env vars -> shared config/credentials (incl. SSO) -> web identity -> container creds -> EC2 IMDSv2`.  The new code checks the credential sources and bails with an error if none are found.

Closes #185

----


nipunn1313:
merged in https://github.com/get-convex/convex-backend/commit/00bd92723422f3bff968230c94ccdeb8c1719832

thanks for the PR

PR #190, opened 2025-08-19
Dont require https for auth provider issuer url

xixixao:
When deploying to kubernetes and using internal URLs we cannot use https. Remove the https check.

This check is here for spec compliance (I think?), removing it does not pose any security risk, as long as people continue configuring https addresses (I don't really see how they would realistically not do it in real-world scenarios, ie when using Clerk et.al.).

cc @gautamg795 

Example of what the URL looks like in our setup: `http://foo-service.bla-namespace.svc.cluster.local:3210/http`

----


thomasballinger:
This is in https://github.com/get-convex/convex-backend/commit/22f22ac6f01e00bb147e7a1cb006e03ae8441c4f, thanks @xixixao

PR #193, opened 2025-08-20
Skip .env.local creation when environment variables exist in process.env

### LINKED ISSUE #191: Convex CLI always creates .env.local even when environment variables are already set
The Convex CLI (`convex dev`) always creates a `.env.local` file and adds it to `.gitignore`, even when the required environment variables (`CONVEX_DEPLOYMENT`, `CONVEX_URL`, etc.) are already available in the shell environment.

This behavior is problematic for users who:
- Use tools like `direnv` with `.envrc` files
- Manage secrets with external tools (1Password, AWS Secrets Manager, etc.)
- Have environment variables set via CI/CD or container orchestration
- Want to avoid leaving any environment settings in local filesystem

## Current Behavior

The CLI creates `.env.local` in these scenarios even when env vars are already set:
1. `convex dev` (with `CONVEX_DEPLOY_KEY` in environment)
2. `convex dev --url "$CONVEX_URL"` (with URL in environment)
3. `convex dev --url "$CONVEX_URL" --admin-key "$CONVEX_ADMIN_KEY"` (with both in environment)

## Expected Behavior

The CLI should detect existing environment variables and skip file creation when:
- `CONVEX_DEPLOYMENT` is already set in the shell environment
- The appropriate URL variable (e.g., `NEXT_PUBLIC_CONVEX_URL`) is already set in the shell environment
- Authentication is available via `CONVEX_DEPLOY_KEY` or CLI arguments

## Reproduction Steps

1. Set environment variables (e.g., via direnv):
   ```bash
   export CONVEX_DEPLOYMENT="dev:amazing-animal-123"
   export CONVEX_URL="https://amazing-animal-123.convex.cloud"
   export NEXT_PUBLIC_CONVEX_URL="https://amazing-animal-123.convex.cloud"
   export CONVEX_DEPLOY_KEY="dev:amazing-animal-123|eyJ..."
   ```

2. Run `convex dev`

3. Observe that `.env.local` is created despite variables being available

## Root Cause Analysis

Looking at the source code in `npm-packages/convex/src/cli/lib/envvars.ts`:

The `envVarWriteConfig` function only checks for environment variables in **files** (`.env.local`, `.env`) but not in the shell environment (`process.env`):

```typescript
// This only reads from files, not process.env
const existingFileContent = ctx.fs.readUtf8File(envFile);
const config = dotenv.parse(existingFileContent);
```

Similarly, `writeDeploymentEnvVar` in `deployment.ts` only checks file-based environment variables.

delkopiso:
## Summary

Fixes issue where Convex CLI always creates `.env.local` files even when required environment variables (`CONVEX_DEPLOYMENT`, `CONVEX_URL`, etc.) are already available in the shell environment. This was problematic for users who manage secrets with external tools, use `direnv`, or deploy in CI/CD environments.

This PR contains two commits that together resolve the issue:

1. **Skip .env.local creation when environment variables exist** - Modified the core logic to check `process.env` before creating `.env.local` files
2. **Fix misleading log messages** - Updated logging to only show file creation messages when files are actually written to disk

## Problem Solved

**Root Cause**: The CLI only checked for environment variables in files (`.env.local`, `.env`) but not in the shell environment (`process.env`).

**Before**: 
- CLI always created `.env.local` even when `CONVEX_DEPLOYMENT` was set via `direnv`, CI/CD, or shell exports
- Users saw misleading "Saved deployment details to .env.local" messages even when no file was needed
- Problematic for users who avoid leaving environment settings in local filesystem

**After**: 
- CLI detects existing environment variables in `process.env` and skips file creation
- Users only see file creation messages when files are actually written
- Respects users' preferred environment management workflows

Fixes #191 
----

### DIFF
--- a/npm-packages/convex/src/cli/lib/deployment.test.ts
+++ b/npm-packages/convex/src/cli/lib/deployment.test.ts
@@ -1,5 +1,14 @@
-import { test, expect } from "vitest";
-import { changesToEnvVarFile, changesToGitIgnore } from "./deployment.js";
+import { test, expect, vi, beforeEach, afterEach } from "vitest";
+import { changesToEnvVarFile, changesToGitIgnore, writeDeploymentEnvVar } from "./deployment.js";
+import { Context } from "../../bundler/context.js";
+
+vi.mock("./envvars.js", async (importOriginal) => {
+  const actual = (await importOriginal()) as any;
+  return {
+    ...actual,
+    gitIgnoreEnvVarFile: vi.fn().mockResolvedValue(false),
+  };
+});
 
 const DEPLOYMENT = {
   team: "snoops",
@@ -82,4 +91,41 @@ test("git ignore changes", () => {
   expect(changesToGitIgnore("!.env.local")).toEqual(
     "!.env.local\n.env.local\n",
   );
+});
+
+const mockContext = {
+  fs: {
+    exists: vi.fn(),
+    readUtf8File: vi.fn(),
+    writeUtf8File: vi.fn(),
+  },
+} as unknown as Context;
+
+const originalProcessEnv = process.env;
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  process.env = { ...originalProcessEnv };
+});
+
+afterEach(() => {
+  process.env = originalProcessEnv;
+});
+
+test("writeDeploymentEnvVar skips file creation when CONVEX_DEPLOYMENT exists with correct value", async () => {
+  process.env.CONVEX_DEPLOYMENT = "prod:tall-bar";
+
+  const result = await writeDeploymentEnvVar(
+    mockContext,
+    "prod",
+    DEPLOYMENT,
+    null,
+  );
+
+  expect(result).toEqual({
+    wroteToGitIgnore: false,
+    changedDeploymentEnvVar: false,
+  });
+
+  expect(mockContext.fs.writeUtf8File).not.toHaveBeenCalled();
 });
--- a/npm-packages/convex/src/cli/lib/deployment.ts
+++ b/npm-packages/convex/src/cli/lib/deployment.ts
@@ -49,6 +49,17 @@ export async function writeDeploymentEnvVar(
   },
   existingValue: string | null,
 ): Promise<{ wroteToGitIgnore: boolean; changedDeploymentEnvVar: boolean }> {
+  const deploymentEnvVarValue =
+    deploymentType + ":" + deployment.deploymentName;
+
+  // Check if the correct value already exists in process.env
+  if (process.env[CONVEX_DEPLOYMENT_ENV_VAR_NAME] === deploymentEnvVarValue) {
+    return {
+      wroteToGitIgnore: false,
+      changedDeploymentEnvVar: false,
+    };
+  }
+
   const existingFile = ctx.fs.exists(ENV_VAR_FILE_PATH)
     ? ctx.fs.readUtf8File(ENV_VAR_FILE_PATH)
     : null;
@@ -57,8 +68,6 @@ export async function writeDeploymentEnvVar(
     deploymentType,
     deployment,
   );
-  const deploymentEnvVarValue =
-    deploymentType + ":" + deployment.deploymentName;
 
   if (changedFile !== null) {
     ctx.fs.writeUtf8File(ENV_VAR_FILE_PATH, changedFile);
--- a/npm-packages/convex/src/cli/lib/envvars.test.ts
+++ b/npm-packages/convex/src/cli/lib/envvars.test.ts
@@ -1,0 +1,66 @@
+import {afterEach, beforeEach, expect, test, vi} from "vitest";
+import {writeConvexUrlToEnvFile} from "./envvars.js";
+import {Context} from "../../bundler/context.js";
+
+vi.mock("./utils/utils.js", () => ({
+	loadPackageJson: vi.fn().mockResolvedValue({name: "test-project"}),
+	ENV_VAR_FILE_PATH: ".env.local",
+}));
+
+const mockContext = {
+	fs: {
+		exists: vi.fn() as any,
+		readUtf8File: vi.fn() as any,
+		writeUtf8File: vi.fn() as any,
+	},
+	crash: vi.fn() as any,
+} as unknown as Context;
+
+const originalProcessEnv = process.env;
+
+beforeEach(() => {
+	vi.clearAllMocks();
+	process.env = {...originalProcessEnv};
+});
+
+afterEach(() => {
+	process.env = originalProcessEnv;
+});
+
+test("writeConvexUrlToEnvFile process.env behavior", async () => {
+	// Test core functionality: skip file creation when env var exists with correct value
+	process.env.CONVEX_URL = "https://test.convex.cloud";
+	(mockContext.fs.exists as any).mockReturnValue(false);
+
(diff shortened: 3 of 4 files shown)


delkopiso:
@nipunn1313 could you take at look at this PR for me? I'd appreciate any feedback

thomasballinger:
@delkopiso these values aren't secret, and they're needed by the frontend to know where to connect to. What's your use case, what's the flow you'd like to support here?

thomasballinger:
I'll wait to hear what you're trying to support, if I were guessing I'd say
- check out [CONVEX_DEPLOY_KEY](https://docs.convex.dev/cli/deploy-key-types)
- are you creating apps that dont' have a frontend, or deploy the frontend seprately? If so, what would help you here?
- is this for self-hosted cases where the deployment name and URL are constant, and you don't want to enable creating new local dev deployments? If so, say more about that

delkopiso:
@thomasballinger my primary usecase is working on separate projects concurrently. I first ran into issues because the access token setup assumes a single account for all projects. Ultimately, there are two things I want to solve for:
- Not having to manage variables (secret or not) in more than one way. My projects aren't using `.env` files.
- Being able to context switch across projects without need to redo the cli auth flow.

I documented in https://github.com/get-convex/convex-backend/issues/191 that setting [CONVEX_DEPLOY_KEY](https://docs.convex.dev/cli/deploy-key-types) didn't prevent the dev flow from trying to write the files.

thomasballinger:
Hm, re multiple accounts that's just not something we're anticipating; can you add your single account to multiple teams instead? It sounds like you want to be logged into multiple accounts and have which account is used depend on the project; nothing like this planned but if there's a good reason you need multiple accounts it sounds like that's something to think about for the future.

Re the first bullet
> Not having to manage variables (secret or not) in more than one way. My projects aren't using .env files.

Every time someone checks your project repo out and runs convex, they get a new CONVEX_URL. It sounds like you want to take this environment variable and move it to another system. You have another way of telling the frontend/client/website what URL to use to talk to the backend.

Are you saying that if the correct environment variables exist in `process.env` (and they match what the CLI *would* have generated) then they shouldn't be written? We could add an option for that. I don't want to silently not write .env.local just because the environment variables are present, because the more important thing is that the website have them available.

I totally agree re CONVEX_DEPLOYMENT, that shouldn't be in .env.local; that's just a convenient place to put it to remember which project a repo checkout is associated with.

Two potential alternative proposals here:
1. How about a dev option for not writing the client environment variable at all? This is the CONVEX_URL / VITE_CONVEX_URL etc. one.
2. For CONVEX_DEPLOYMENT, it would be nice if CONVEX_DEPLOY_KEY or `--env-file` also caused this not to be written. It sounds like you want another option, which is if CONVEX_DEPLOYMENT is set (and as long as that's a valid) it won't be written. That part sounds good.

delkopiso:
@thomasballinger that's correct on all counts. For CONVEX_DEPLOYMENT, CONVEX_DEPLOY_KEY not writing the file would suffice too.

suiramdev:
I also need it because I have configured convex in a “packages/convex” directory inside my turbo repository, where all applications and packages depend on a single “.env” file at the root using “dotenv-run.”

thomasballinger:
@suiramdev do you not use any .env.local files, or similar file that is not committed to the repo?

Each time you change dev deployments, (in the future, you might use different deployments for different branches) we need to update the NEXT_PUBLIC_CONVEX_URL / VITE_CONVEX_URL / CONVEX_URL so the frontend knows where to find that deployment. Where would you want this to be written to? This url should be in version control.

@delkopiso Similar question: where should this be written?

We can eventually move away from CONVEX_DEPLOYMENT by moving it to a new hidden file that should be git-ignored like `.convex/config.json`. But whenever the deployment URL changes you need to know what url to use to find it.

suiramdev:
My intention in responding to this PR was to 'bump' it. Let me explain why I am interested in it more clearly @thomasballinger. Despite a small example showcasing its use with Turborepo, Convex appears to have been designed to attach to one front end rather than multiple ones via a monorepository. In the context of my boilerplate, I would like to use Convex as a shared package in the `@packages/convex` directory, which is powered by Turborepo. For the sake of ease of setup, I have chosen to maintain one and only one `.env` file [at the root of this monorepository](https://github.com/suiramdev/next-starter/blob/add-convex/.env.example).

However, I am unable to specify an environment file or use the one provided through process.env, as I use dotenv-run to spread that root `.env` file across my packages and apps.
This results in the creation of a `.env.local` file in the `@packages/convex` directory each time `pnpx convex dev` is run, and requires the user to interact with the CLI interface to choose their convex environment.

In my opinion, interaction with the CLI should be optional. It should be similar to what some CLI tools do with the option `-i`.
Convex should use the environment variable whenever it already exists.

Am I missing something that convex is trying to solve in this generated `.env.local` file, which in our case conflicts with owns? Shouldn't the environment variables defined by convex with `pnpx convex env set` be located in a very specific convex folder, as you said `.convex/`?

Let me know if you'd like more information, or if I'm missing out some details.

---

I discovered that convex exposes a `--env-file` option on `pnpx convex dev` but it does not appear to be documented and does not work in the same way we intend here : [A user also questioned about it](https://discord.com/channels/1019350475847499849/1399819522139099218)

thomasballinger:
@suiramdev I see you have a .env at the root, is that .env version controlled? If not, great, let's make `convex dev` write CONVEX_URL and CONVEX_DEPLOYMENT there. If so, would you also want a .env.local there too?

> Am I missing something that convex is trying to solve in this generated .env.local file

It sounds like you get it, but to reiterate, whenever `convex dev` runs it needs to set the CONVEX_URL / VITE_CONVEX_URL / NEXT_PUBLIC_CONVEX_URL to the url of the (possibly newly created) deployment. This can absolutely be moved to a different .env file!

thomasballinger:
> Shouldn't the environment variables defined by convex with pnpx convex env set be located in a very specific convex folder, as you said .convex/?

This is pretty different, `convex env set` is for setting environment variables *in the deployment*. The environment variables we're talking about here are not in the deployment, they're for use in the applications that need to connect to the deployment.

thomasballinger:
> In my opinion, interaction with the CLI should be optional.

Absolutely, this is important. The only guaranteed way to make `convex dev` non-interactive today is to 1) use a CONVEX_DEPLOY_KEY (this is the suggested way to do build pipelines in e.g. Vercel) or 2) have a configured CONVEX_DEPLOYMENT, or 3) run `npx convex dev --configure` with a bunch of additional flags like `--team` and `--project`.

thomasballinger:
> However, I am unable to specify an environment file or use the one provided through process.env, as I use dotenv-run to spread that root .env file across my packages and apps.

Agree --env-file is not fully implemented yet, it should allow you to save CONVEX_DEPLOYMENT and CONVEX_URL / VITE_CONVEX_URL / NEXT_PUBLIC_CONVEX_URL to a different file.

> use the one provided through process.env

What do you mean by this?

thomasballinger:
Would love to hear more, so far I'm hearing

1. Let's provide a way to specify a different file than .env.local (e.g. ../../.env) to write CONVEX_URL to
2. (lower priority) provide to disable writing to this file, instead crashing if CONVEX_URL doesn't match
3. (lower priority) write CONVEX_DEPLOYMENT to a different place

let me know how these differ from your expectations

thomasballinger:
I broke 1) out into https://github.com/get-convex/convex-backend/issues/206

Issue #194, opened 2025-08-20
DISABLE_BEACON is not obeyed by the backend


xixixao:
DISABLE_BEACON is only obeyed by the docker compose setup, not by the backend itself. I guess the error message might be hinting at this with:

```
You can disable this telemetry by setting the --disable-beacon flag or the DISABLE_BEACON environment variable if you are self-hosting using the Docker image.
```

but it's confusing (like what is "the Docker image" in that sentence?). It'd be better if the backend binary itself read the value from the env var.

nipunn1313:
yep! Agreed! Little bit confusing.
If someone wants to grab this one - it seems pretty doable. Have both the env var and the command line flag - if either is set, then have the backend obey.

nipunn1313:
Fixed - thanks @STAR-173

PR #196, opened 2025-08-21
Fix: Respect `DISABLE_BEACON` env var in backend

STAR-173:
This change addresses issue #194  by making the backend binary directly obey the DISABLE_BEACON environment variable, in addition to the existing --disable-beacon flag.

Previously, the environment variable was only handled by the Docker Compose setup, which was confusing for users not using Docker.

This PR also clarifies the startup message to reflect that either the environment variable or the command-line flag can be used to disable telemetry.

----

### DIFF
--- a/crates/local_backend/src/config.rs
+++ b/crates/local_backend/src/config.rs
@@ -103,7 +103,7 @@ pub struct LocalConfig {
     /// self-hosted Convex will periodically communicate with a remote beacon
     /// server. This is to help Convex understand and improve the product.
     /// If set, the self-host beacon will not be sent.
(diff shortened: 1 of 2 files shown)


nipunn1313 (approved):

nipunn1313:
merged in https://github.com/get-convex/convex-backend/commit/ad4e3eec7beeec2a7b04a4bddad5e2cf5c06814d

I made a couple modifications - used the boolish value parser so that it supports `1` or `true` as true values.

nipunn1313:
Thanks for the contribution!!

PR #202, opened 2025-08-26
fix: playground docs

mrpaaradox:
Few weeks ago there was a change in the Playground docs where instead of (agent-playground) only (agent) was selected to import the definePlaygroundAPI. Previous version of the docs was working fine. Hence I am making this PR to reverse that change. 

I have tested agent package & I couldn't find any supporting changing which can help us import the definePlaygroundAPI.
Also I see no updates for the same on the Official Convex Playground page as shown below.

Screenshot 2025-08-26 at 6 02 39 PM

----

### DIFF
--- a/npm-packages/docs/docs/agents/playground.mdx
+++ b/npm-packages/docs/docs/agents/playground.mdx
@@ -24,13 +24,17 @@ There is also a [hosted version here](https://get-convex.github.io/agent/).
 **Note**: You must already have a Convex project set up with the Agent. See the
 [docs](./getting-started.mdx) for setup instructions.
 
+```sh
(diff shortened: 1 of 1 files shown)


nipunn1313:
@ianmacartney - can you review? If it looks good, I can merge.

ianmacartney:
@mrpaaradox as of version 0.1.17 (I believe) the Agent exports the `definePlaygroundAPI` function directly and is generally the pattern I recommend. That's a good callout that the hosted playground is outdated - I'm making that change to update it there to the new syntax. If you have any issues importing it from the Agent directly please let me know - filing a GitHub issue in the Agent repo would be preferred: https://github.com/get-convex/agent

PR #203, opened 2025-08-26
React: Optionally treat loading as unauthenticated

z3ttee:
Hey all, I used convex for some weeks now and came across the `<Unauthenticated />` react component.
I noticed when using SSR in NextJS there is a short time window where the page renders nothing when using said component together with `<Authenticated />`. I looked into the code and found out, that when the authentication state is loading, neither of said components render anything.

To have more control over this and to fix this small issue I want to introduce a prop on the `<Unauthenticated />` component.

How does it work?
With the prop `loadingEqualsUnauthenticated` developers can decide wether they want to treat the loading state as unauthenticated or not. So when this gets set to `true`, the loading state is ignored and the children are rendered to the DOM. 
This option defaults to `false`, so this should not be a breaking change. This is something similar to what Clerk provides through their built-in components: [loadingEqualsUnauthenticated](https://clerk.com/docs/nextjs/components/control/signed-in#properties)

Please let me know what you think of this change. Any feedback is welcome.

----

### DIFF
--- a/npm-packages/convex/src/react/auth_helpers.test.tsx
+++ b/npm-packages/convex/src/react/auth_helpers.test.tsx
@@ -1,9 +1,22 @@
 /**
  * @vitest-environment custom-vitest-environment.ts
  */
-import { test } from "vitest";
+import { render, screen } from "@testing-library/react";
 import React from "react";
+import { describe, expect, test } from "vitest";
 import { Authenticated, AuthLoading, Unauthenticated } from "./auth_helpers.js";
+import { ConvexReactClient } from "./client.js";
+import { ConvexProviderWithAuth } from "./index.js";
+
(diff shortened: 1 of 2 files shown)


nipunn1313:
Seems like a reasonable desire, but we also want to balance with the desire for simplicity - additional flags -> additional complexity. We'll have to noodle on this to decide if it's worth.

thomasballinger:
Note that you can accomplish this today with a hook, or by rendering the same thing in `<Unauthenticated>` or `<Loading>`! How about instead of a flag, just a new component called `<UnauthenticatedOrLoading>`?

z3ttee:
I used the hook for checking the loading state before, but for me it felt weird having components that render on a given auth state but both cannot handle loading states. I think having a dedicated component for this is a good idea instead of having a flag

PR #208, opened 2025-08-31
Clarify documentation regarding instance names / connection strings

moritzuehling:
Clarifies the documentation regarding the instance name and db connection strings.

---

There's more detail to cover outside of `--instance-name`, e.g. I've entirely ignored `multi-schema` - a rather useful abstraction layer to quickly bring test envrionments up and down. 

Alas, I couldn't run prettier on this because I'm using the github webui for this PR. I might come back tomorrow to fix it and/or address feedback.

I'm neither married to the wording nor the location of these settings, but because they're (atm) entirely undocumented and I could only find out the behavior after reading through your rust code for a while, I think it makes sense to have some config here. 

Perhaps it might make sense to create a longer, more detailed markdown file that goes into the specifics. At the same time, it would be useful to expose the the `--db` flag directly via the docker image. 

Anyway, I hope this contribution is useful - at least to inspire contributions by the core maintainers :)

----

### DIFF
--- a/self-hosted/README.md
+++ b/self-hosted/README.md
@@ -327,6 +327,18 @@ export MYSQL_URL=mysql://<your-username>:<your-password>@aws.connect.psdb.cloud
 docker compose up
 ```
 
+### Database Names
+
+The database name that the Convex backend uses is equivalent to the instance name (replacing `-` with `_`). If no instance name is set, the Docker image defaults to `convex-self-hosted`, and the Convex backend will connect to the database `convex_self_hosted`.
+
+For the docker container, you can set the instance name via the `INSTANCE_NAME` envrionment variable.
+
+```sh
+export POSTGRES_URL='<connection string>'
+export INSTANCE_NAME='your-instance-name'
+psql $POSTGRES_URL -c "CREATE DATABASE your_instance_name;"
+```
+
 ## Using S3 Storage
 
 By default, the backend stores file data on the filesystem within the docker


moritzuehling:
Also, hello convex team - this is unrelated, but I think this code can never be reached:

https://github.com/get-convex/convex-backend/blob/fde777b69781d15c3c67eb2dc61549968f9150b5/crates/clusters/src/lib.rs#L65

-  Because`cluster_url.query_pairs()` must be empty
- Because `anyhow::ensure!(cluster_url.query().is_none(), /* ... */` [here](https://github.com/get-convex/convex-backend/blob/fde777b69781d15c3c67eb2dc61549968f9150b5/crates/clusters/src/lib.rs#L33-L37)

==> So, if this collection ever was non-empty, we would have error'd out earlier.

self-hosted/README.md
```diff
@@ -327,6 +327,18 @@ export MYSQL_URL=mysql://<your-username>:<your-password>@aws.connect.psdb.cloud
 docker compose up
 ```
 
+### Database Names
+
+The database name that the Convex backend uses is equivalent to the instance name (replacing `-` with `_`). If no instance name is set, the Docker image defaults to `convex-self-hosted`, and the Convex backend will connect to the database `convex_self_hosted`.
+
+For the docker container, you can set the instance name via the `INSTANCE_NAME` envrionment variable.
+
```
nipunn1313, line 335: looks good. I am going to change the wording to be very clear that this is an example - since your example is specific to postgres, but the concept carries across any underlying persistence database.

nipunn1313 (approved):

nipunn1313:
merged in efed817a4b07cbf27d94112083452b4e09f10801

PR #214, opened 2025-09-05
Feature stdin env cli

graham:
Setting JWTs from the command line is a little tedious and difficult to get right (specifically because they are multi-line files). This allows the user to echo/pipe/etc the value into the value into the cli.

`cat keys/my-private-key.txt | npx convex env set JWT_PRIVATE_KEY`

----


thomasballinger:
this is in https://github.com/get-convex/convex-backend/commit/ef8aee4643b89ff452cee427d9b7c0078249ab7e, thanks!

PR #220, opened 2025-09-15
DOCS: Improved Workos Convex Provider to work with org switching

jappi00:
Hello everyone,

I encountered some issues with the provider from the documentation. I found an updated version directly from WorkOS here: https://github.com/workos/template-convex-nextjs-authkit/blob/main/components/ConvexClientProvider.tsx

I'm not entirely sure if I did this PR correctly! 😅

----

### DIFF
--- a/npm-packages/docs/docs/auth/authkit.mdx
+++ b/npm-packages/docs/docs/auth/authkit.mdx
@@ -432,14 +432,14 @@ have a working Next.js app with Convex. If not follow the
 
     ```tsx title="components/ConvexClientProvider.tsx"
     'use client';
-
-    import { ReactNode, useCallback, useRef } from 'react';
+    
+    import { ReactNode, useCallback } from 'react';
     import { ConvexReactClient } from 'convex/react';
     import { ConvexProviderWithAuth } from 'convex/react';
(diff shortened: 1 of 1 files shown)


thomasballinger:
Thanks @jappi00 , this looks reasonable to me! I want to hold off because we're actively making changes to https://github.com/workos/template-convex-nextjs-authkit still; we'll get this change (or something similar) in soon. Do you see org switching working with this code but not with the current implementation?

jappi00:
thanks for the quick answer @thomasballinger!

Yes, I only get it to work with the updated Provider:

With the old one it only works after a refresh of the page (before the backend auth does not get refreshed.)

That is how I switch the organizations:
```js
import { useAuth } from "@workos-inc/authkit-nextjs/components";

// Switch organization context
await refreshAuth({ organizationId: team.workos_id });
```

Once I've polished the complete implementation, I'd like to extend the documentation to cover organization switching and user syncing, unless you have any objections.

jappi00:
For more Context: https://github.com/workos/template-convex-nextjs-authkit/commit/afd809700f2460cf3a7c95ca0f47f37ab5606d0c

PR #240, opened 2025-10-15
Log Streaming Webhook sink: add basic auth

leventov:
Add Basic HTTP auth to webhook logs sink.

The main objective is streaming logs to ClickHouse: https://clickhouse.com/docs/interfaces/http

Cf. this Discord discussion: https://discord.com/channels/1019350475847499849/1296860730104217610

P.S. This patch is vibe-coded, I'm not an expert in Rust. I didn't test this patch.

----

### DIFF
--- a/crates/local_backend/src/log_sinks.rs
+++ b/crates/local_backend/src/log_sinks.rs
@@ -92,6 +92,7 @@ pub async fn add_datadog_sink(
 pub struct WebhookSinkPostArgs {
     url: String,
     format: WebhookFormat,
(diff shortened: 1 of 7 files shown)


nipunn1313:
I don't think basic auth adds much value over putting a token in the URL as a query parameter.
I would recommend you do that instead.

The discord discussion you linked mentions a concern about security. What is the threat model of security you are worried about? URL query parameters are encrypted on HTTPS. Basic auth has similar security properties.

Open to different proposals if there's a need.

Issue #102, opened 2025-05-22
Feature Proposal: Allow escaped folders inside


wkd-kapsule:
Hi,

First of all, huge thanks to anyone involved in this project. This is a real gem and a pleasure to use.

The feature I'd like to see added to Convex is a way to escape folders inside the `convex` folder. In order to have more control on their organization and the `api` generated.

Let's say I have a `member` table and I'd like it's `api` to be:
`api.member.crud.create`
`api.member.crud.create`
`api.member.crud.create`
`api.member.crud.create`

``` md
convex
├─ member
│ ├─ crud.ts
│ └─ checkout.ts
├─ member.ts
```

nipunn1313:
I *think* you can use `index.ts` (eg user/index.ts) to get the API you're looking for.
Try it?

wkd-kapsule:
> I _think_ you can use `index.ts` (eg user/index.ts) to get the API you're looking for.
> Try it?

I did and unfortunately it doesn't work. It generates `api.user.index.*`.
But still, even if it worked it wouldn't give as much customization as my proposal. All `api.user.*` would have to live in `index.ts`.

nipunn1313:
ah yeah - you're right. Sorry for the misleading suggestion.

Yeah - this feature idea is good. Either via an `index.ts` or doing something similar to next.
We'll keep in in mind for going forward. @ianmacartney has been making a list of ideas for a 2.0 of Convex - and something in this space seems reasonable.

wkd-kapsule:
I'm glad that you like the proposal. It will be cool to see it in a future version. Maybe even before 2.0? It doesn't seem like a difficult problem to tackle or involving a breaking change.

Anyway, if and whenever you implement this, please consider my version of it instead of the one using `index.ts`. As mentioned, the latter would solve some cases but leave other drawbacks. If your team takes the time to add the feature, they might as well go for the fully customizable version rather than something that people will eventually complain about again.

ianmacartney:
Here's the way I think about it:

Supporting `index.ts` would be convenient, I agree. It looks a bit ugly/verbose, and it's hard to migrate from having `users.ts` to `/users` without a breaking change.

I see folder naming as an intent to organize code differently than where you expose endpoints. This is already possible today! You can put code in a `lib/checkout/` folder, then re-export them from a top-level /orders.ts. You can decide whether to export it in both locations, or only in one. 
Here's a repo I just made to show how to do it: https://github.com/ianmacartney/code-organization-pattern

[...]

PR #246, opened 2025-11-02
deduplicate the filename when is a single leaf and has the same name as the file

sebas5384:
### Problem
Implements a namespace deduplication for Convex API code generation. When organizing Convex functions (queries, mutations, actions) into separate files, the generated API namespace no longer duplicates the function name if it matches the filename and the file contains a single function.

Currently organizing a single function in a dedicated file would create redundant namespaces:

```js
// File: convex/blog/post/getComments.ts
export const getComments = query(...)

// Generated namespace (BEFORE):
api.blog.post.getComments.getComments  ❌ (duplicated)

// Generated namespace (AFTER):
api.blog.post.getComments  ✅ (clean)
```

### Benefits
🎯 Cleaner API Namespaces: Eliminates redundancy for the common pattern of single-function-per-file organization
🛡️ Safe: Zero breaking changes - only affects new single-function-per-file pattern

Issue related: https://github.com/get-convex/convex-backend/issues/102#issue-3083341970

### DIFF
--- a/npm-packages/convex/src/cli/codegen_templates/component_api.test.ts
+++ b/npm-packages/convex/src/cli/codegen_templates/component_api.test.ts
@@ -1,0 +1,100 @@
+import { test, expect } from "vitest";
+import type {
+  AnalyzedFunction,
+  AnalyzedModule,
+} from "../lib/deployApi/modules.js";
+import type { CanonicalizedModulePath } from "../lib/deployApi/paths.js";
+
+const mockCtx = {
+  crash: async (error: any) => {
+    throw new Error(error.printedMessage);
+  },
+} as any;
+
+function createMockFunction(name: string): AnalyzedFunction {
+  return {
+    name,
+    udfType: "Query",
+    visibility: { kind: "public" as const },
+    args: "{}",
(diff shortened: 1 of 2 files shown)


nipunn1313:
I don't believe this is backwards compatible. Can't take this PR since it's making some fundamental changes to the API interface, but the input is taken. The function call syntax can be verbose in files with a single function. (CC @ianmacartney / @thomasballinger )

Would recommend you use default exports in the meantime.

sebas5384:
hey @nipunn1313 thanks for looking into this PR, could you tell me more what fundamental changes are you referring to? I'll use the default export

sebas5384:
@nipunn1313 just tested using the default export and it adds a "default" at the end.

Is there a way of avoiding this? I could work on something to at least remove the default last part?

nipunn1313:
you're welcome to fork the library if you want.

It's just that the change is not easy to do backwards compatibly, so we're unlikely to roll it out incrementally like this. Perhaps something like it could be rolled into a larger v2.0 release of an API - but would want our team to think about it.

PR #248, opened 2025-11-05
Add optional `name` parameter to `v.object()` for cross-language type safety

dickermoshe:
## Summary

This PR adds an optional `name` parameter to the `v.object()` validator constructor to enable better type generation for Convex clients in languages other than TypeScript.

## Problem

When generating type-safe clients for languages without TypeScript's powerful type system (like Python, Java, Go, etc.), we need consistent, reusable type names for object schemas. Currently, without named objects, client generators must create arbitrary names like `Object123` or `AnonymousObject`, making types non-reusable and harder to work with. These types are also not the same across different requests. So 2 functions which return the identical object, return different classes.

## Solution

Add an optional `name` parameter to `v.object()` that allows developers to specify meaningful names for their object types:

```typescript
// Before
const userSchema = v.object({
  name: v.string(),
  age: v.number(),
});

// After
const userSchema = v.object({
  name: v.string(),
  age: v.number(),
}, { name: "User" });
```

This name would then appear in the output of the `funtion-spec` command. Client library authors could then use this command to generate data models like:
```python
class User(BaseModel):
  name: str
  age: float
```
which could be used across multiple different requests.

The alternative would be:
```python
class ConvexModel835989(BaseModel):
  name: str
  age: float
```

While TypeScript already handles types internally, providing explicit names could also benefit code generation and type declarations for TypeScript itself. The primary motivation, however, is to support non-TypeScript languages.

## Backward Compatibility

This change is fully backward compatible:
- Existing code using `v.object(fields)` continues to work unchanged
- The `name` parameter is optional and defaults to `undefined`
- No breaking changes to the API surface

### DIFF
--- a/npm-packages/convex/src/values/validator.ts
+++ b/npm-packages/convex/src/values/validator.ts
@@ -165,11 +165,14 @@ export const v = {
 
   /**
    * Validates that the value is an Object with the given properties.
-   * @param fields An object specifying the validator for each property.
-   * @param name An optional name for this object validator.
+   * @param options Optional configuration for this object validator.
+   * @param options.name An optional name for this object validator.
    */
-  object: <T extends PropertyValidators>(fields: T, name?: string|undefined) => {
-    return new VObject<ObjectType<T>, T>({ isOptional: "required", fields, name });
+  object: <T extends PropertyValidators>(
+    fields: T,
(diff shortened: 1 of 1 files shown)


dickermoshe:
There should probably be a bigger discussion before this gets merged.

PR #282, opened 2025-12-19
Fix npm create command for npm 11 compatibility

sethconvex:
npm 11 broke passing through config flags with npm create.
Replaced 'npm create convex@latest --' with 'npx create-convex@latest'
and 'npm create convex --' with 'npx create-convex'.


sethconvex:
closing as this will be addressed via a different route.

Issue #288, opened 2025-12-23
Respect port settings for local deployments


kellykampen:
For local deployments it would be very helpful if convex respected the port settings in `.env.local` file and used it to when spinning up the local convex server. Currently it always overwrites with the port being `3210`

This would be very helpful when running multiple worktrees on the same machine. 

In addition reading the `NEXT_PUBLIC_CONVEX_SITE_URL` port would be needed. 

Example `.env.local`
```
NEXT_PUBLIC_CONVEX_URL=http://127.0.0.1:3220
NEXT_PUBLIC_CONVEX_SITE_URL=http://127.0.0.1:3222
```

nipunn1313:
yep seems reasonable!

yash-sangwan:
@nipunn1313 If no one is currently working on this, can I start working on this issue?

dowski:
I was reading the associated PR discussion, and wanted to clarify here:

In addition to being able to run multiple instances of the same app code, do you also want them to share the local Convex backend instance state?

It seems like it's currently possible to run multiple checkouts of a Convex app so long as they're using different `CONVEX_DEPLOYMENT` names which map to backend state.

I locally cloned a sample app and ran it with `CONVEX_AGENT_MODE=anonymous npx convex dev` and it created an `.env.local` with:

```
CONVEX_DEPLOYMENT=anonymous:anonymous-agent

VITE_CONVEX_URL=http://127.0.0.1:3210
```

I ran the same command in another clone of the app and it created an `.env.local` with:

```
CONVEX_DEPLOYMENT=anonymous:anonymous-agent-1

VITE_CONVEX_URL=http://127.0.0.1:3212
```

If they're using the same `CONVEX_DEPLOYMENT` value, that doesn't work.

yash-sangwan:
@dowski You're correct that Convex already handles automatic port assignment when using different CONVEX_DEPLOYMENT values.

This [PR](https://github.com/get-convex/convex-backend/pull/291) addresses a different use case, when developers want manual control over port assignments regardless of deployment naming.

the-kro:
I have a monorepo with components that use other components. trying to start all the dev servers from root at the same time causes conflicts. I work around this now by using sleeps to stagger the start time, but if the existing .env.local's port was used, this wouldn't be a problem.

ianmacartney:
Have you tried setting the ports in the local deployment's config.json to
something unique?

the-kro:
[...]

PR #291, opened 2025-12-26
Fix: Respect port settings from .env.local for local deployments

yash-sangwan:
### Problem #288 
Local deployments always use port 3210/3211, ignoring ports specified in `.env.local`. This prevents running multiple worktrees simultaneously.

### Solution
Read port settings from framework-specific environment variables (NEXT_PUBLIC_CONVEX_URL, VITE_CONVEX_URL, etc.) in .env.local when starting local Convex deployments.

This allows running multiple worktrees on the same machine by specifying different ports in each worktree's .env.local file.

Changes:
- Add getPortsFromEnvFile() to extract ports from .env.local
- Update handleLocalDeployment() to use ports from env vars
- Update handleOffline() to use ports from env vars
- Update handleAnonymousDeployment() to use ports from env vars

Example .env.local:
  NEXT_PUBLIC_CONVEX_URL=http://127.0.0.1:3220 
  NEXT_PUBLIC_CONVEX_SITE_URL=http://127.0.0.1:3222

Backward compatible: defaults to 3210/3211 if not specified. CLI options take precedence over .env.local settings.

Fixes: Multiple worktrees can now run simultaneously without port conflicts

----

### DIFF
--- a/npm-packages/convex/src/cli/lib/localDeployment/anonymous.ts
+++ b/npm-packages/convex/src/cli/lib/localDeployment/anonymous.ts
@@ -33,6 +33,7 @@ import {
   isOffline,
   generateInstanceSecret,
   choosePorts,
+  getPortsFromEnvFile,
   LOCAL_BACKEND_INSTANCE_SECRET,
 } from "./utils.js";
 import { handleDashboard } from "./dashboard.js";
@@ -141,10 +142,16 @@ export async function handleAnonymousDeployment(
     adminKey = data.adminKey;
   }
 
+  // Read port settings from .env.local if not explicitly provided via CLI options
+  const portsFromEnv = options.ports ? { cloud: null, site: null } : await getPortsFromEnvFile(ctx);
+
   const [cloudPort, sitePort] = await choosePorts(ctx, {
     count: 2,
     startPort: 3210,
-    requestedPorts: [options.ports?.cloud ?? null, options.ports?.site ?? null],
+    requestedPorts: [
+      options.ports?.cloud ?? portsFromEnv.cloud ?? null,
+      options.ports?.site ?? portsFromEnv.site ?? null,
+    ],
   });
   const onActivity = async (isOffline: boolean, _wasOffline: boolean) => {
     await ensureBackendRunning(ctx, {
--- a/npm-packages/convex/src/cli/lib/localDeployment/localDeployment.ts
+++ b/npm-packages/convex/src/cli/lib/localDeployment/localDeployment.ts
@@ -23,6 +23,7 @@ import { promptSearch } from "../utils/prompts.js";
 import { LocalDeploymentError, printLocalDeploymentOnError } from "./errors.js";
 import {
   choosePorts,
+  getPortsFromEnvFile,
   printLocalDeploymentWelcomeMessage,
   isOffline,
   LOCAL_BACKEND_INSTANCE_SECRET,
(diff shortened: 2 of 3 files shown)


nilskroe:
Is this really not possible... Wasted probably 3 hours now first trying preview deployments but they only support deploying and no hot reload, then not local deployment, docker etc. and everything is fucked at some point and not working

yash-sangwan:
Could you explain more, I'll try to fix that if that possible

nilskroe:
I didn't test yet your PR, sounds really promising. Tried only every "native" way with convex so far

yash-sangwan:
Makes sense, thanks for clarifying.

If you get a chance to try it and something feels off (or if it helps your setup), would love to hear either way. Otherwise, we can wait for maintainer review.

nilskroe:
Looks promising, but this basically only fixes that you can run different projects locally at the same time on different ports right? Nice would be same project multiple local instances at the same time or is that also doable?

nilskroe:
Ah wild, it does actually work with multiple local deployments for same project but the dashboard ui is crashing from convex "We encountered an error loading this page.

Please try again or contact us at [support@convex.dev](mailto:support@convex.dev) for support with this issue.

Event ID: 2b7d2abe5bfb44bcadf1eb17ff0d00ca
[Convex Status page](https://status.convex.dev/)"

nilskroe:
Screenshot 2025-12-29 at 12 58 00 AM
Damn perfect, exactly what I needed. 🚀

yash-sangwan:
That’s great! Happy it helped. Let me know if you find any issues, please describe the issue clearly, and I’ll try to fix it.

nilskroe:
Depends: Do you intend it so people can run multiple local deployments for the same project in parallel on different ports or just so you can run multiple local deployments but each for a different project in parallel. Second thing is covered, for the first thing are a lot more changes required

yash-sangwan:
This PR is intended to solve [this](https://github.com/get-convex/convex-backend/issues/288) issue.  
For the first one (people can run multiple local deployments for the same project in parallel on different port), you can raise a new issue and see if any maintainer approves.

kellykampen:
@yash-sangwan This is fantastic, wonderful holiday gift. 

In my orignal issue I should have been more clear.  I was indeed hoping that it would be able to run multiple local instances for the same project in parallel. 

I will wait for this to be merged and raise another issue for parallelization of local deployments.

yash-sangwan:
Totally makes sense, thanks for clarifying. let’s take the parallel local instances part as a follow-up issue after this gets merged.

ya7010:
This is exactly the feature I was looking for. Is there a merge plan?

ianmacartney (commented):
FYI there was recently added a path that sets the *CONVEX_SITE_URL - we should re-use the code for that naming if possible

arielweinberger:
Worktrees are quite common these days with coding agents. Cannot have multiple agents in parallel working on different worktrees, each with their own Convex devserver, because the worktree setup scripts cause a conflict due to the default port being already taken. Causes unnecessary friction. Can this be merged? @ianmacartney 

The only alternative atm is to use docker containers and map each container to a different host port. Not very pretty, and not straightforward in terms of livereload/ergonomics.

arielweinberger:
Just a thought @yash-sangwan @ianmacartney if I understand correctly, in the case of anonymous.ts, it would still try to extract the Convex backend ports from the env file (.env.local)? I suggest checking `process.env` first. Many tools (e.g. Conductor) allow you to configure worktree startup using env var override as a mechanism. As with 99% of developer tooling out there, if you run a command with an env variable defined in your shell, it's expected that it'll take precedence over any `.env*` file in the filesystem.

Please let me know if I misunderstood the implementation. Sorry if that's the case.

akd-io:
Would love to see this merged. Can I help with anything? ❤️

PR #295, opened 2025-12-28
Fix: Preserve numeric literal types in function spec

### LINKED ISSUE #213: `npx convex function-spec` Reports incorrect literal
The following task:
```ts
export const query10 = query({
  args: {
    i: v.literal(1),
  },
  returns: {
    i: v.literal(1),
  },
  handler: (ctx, args) => {
    return args;
  },
});
```
Will produce the following schema:
```
{
      "args": {
        "type": "object",
        "value": {
          "i": {
            "fieldType": {
              "type": "literal",
              "value": 1
            },
            "optional": false
          }
        }
      },
      "functionType": "Query",
      "identifier": "tasks.js:query10",
      "returns": {
        "type": "object",
        "value": {
          "i": {
            "fieldType": {
              "type": "literal",
              "value": 1
            },
            "optional": false
          }
        }
      },
      "visibility": {
        "kind": "public"
      }
    }
```
However this is incorrect.
The literal type should be `1.0`.
This results in a crash at runtime:
```
ClientError.serverError(msg: [Request ID: 67be452c7b31cf99] Server Error
  ArgumentValidationError: `1` does not match literal validator `v.literal(1.0)`.Path: .i
  )
```
It's not enough just to copy the contents of the literal to the function-spec, we should really have another field called `literalType` for this.

yash-sangwan:
## Problem

When using `v.literal(1)` in TypeScript, the function-spec incorrectly reported the literal value as `1.0` instead of `1`, causing runtime validation errors:

```typescript
export const query10 = query({
  args: {
    i: v.literal(1),
  },
  returns: {
    i: v.literal(1),
  },
  handler: (ctx, args) => {
    return args;
  },
});
```

This produced a function-spec with `value: 1.0`, but when calling the function with `1`, it failed with:

```
ClientError.serverError(msg: [Request ID: 67be452c7b31cf99] Server Error
  ArgumentValidationError: `1` does not match literal validator `v.literal(1.0)`.Path: .i
)
```

### Root Cause

JavaScript doesn't distinguish between integers and floats - both `1` and `1.0` are the same `number` type. When `v.literal(1)` was serialized:

1. TypeScript's `convexToJson(1)` treated it as a regular number
2. JSON serialization couldn't preserve whether the user intended Int64 or Float64
3. Backend deserialized the JSON number as Float64 by default
4. Function-spec showed `1.0` instead of `1`
5. Runtime validation failed because the stored literal type didn't match the input

## Solution

Added an optional `literalType` field to explicitly encode the intended numeric type:

- `v.literal(1)` → `{ type: "literal", value: 1, literalType: "Int64" }`
- `v.literal(1.5)` → `{ type: "literal", value: 1.5, literalType: "Float64" }`
- `v.literal("test")` → `{ type: "literal", value: "test", literalType: "String" }`
- `v.literal(true)` → `{ type: "literal", value: true, literalType: "Boolean" }`

### TypeScript (`npm-packages/convex/`)

**`src/values/validators.ts`**
- Added `literalType?: "Float64" | "Int64" | "Boolean" | "String"` to `ValidatorJSON` type
- Updated `VLiteral.json` getter to detect and include the appropriate `literalType`
- Integer detection uses `Number.isInteger()` for JavaScript numbers

**`src/cli/lib/deployApi/validator.ts`**
- Updated zod schema to include optional `literalType` field

**`src/values/validator.test.ts`**
- Added tests verifying each literal type includes correct `literalType` field

### Rust (`crates/common/`)

**`src/schemas/json.rs`**
- Added `literal_type: Option<String>` to `ValidatorJson::Literal` variant
- Fixed critical bug: `LiteralValidator::Int64` now serializes as plain JSON number instead of `{"$integer": "..."}`
- Added deserialization logic to use `literalType` hint when converting JSON to validators
- Added 4 new tests for literal type preservation and roundtrip conversion

### Integration Tests

**`npm-packages/udf-tests/convex/args_validation.ts`**
- Added test functions for `v.literal(1)`, `v.literal(1.5)`, `v.literal("test")`, and `v.literal(true)`

## Testing

- All TypeScript tests pass (28/28)
- TypeScript build successful
- Manual verification confirms correct JSON output

**Before:**
```json
{
  "type": "literal",
  "value": 1.0
}
```

**After:**
```json
{
  "type": "literal",
  "value": 1,
  "literalType": "Int64"
}
```

## Backward Compatibility

- `literalType` field is optional and uses `#[serde(skip_serializing_if = "Option::is_none")]`
- Old validators without `literalType` still work (default to inferring type from value)
- No migration needed for existing deployments

## Fixes

Closes #213

----

### DIFF
--- a/crates/common/src/schemas/json.rs
+++ b/crates/common/src/schemas/json.rs
@@ -646,6 +646,8 @@ pub enum ValidatorJson {
     Any,
     Literal {
         value: JsonValue,
+        #[serde(skip_serializing_if = "Option::is_none")]
+        literal_type: Option<String>,
     },
     #[serde(rename_all = "camelCase")]
     Id {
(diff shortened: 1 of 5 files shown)


ianmacartney (changes requested):
I would rather Int64 be only used for bigint / v.int64 types (`v.literal(1n)`) - otherwise the api spec would be ambiguous which fields accept/return bigint vs. number if the field type says Int64.
My guess is you're using a client which is encoding an int64 as the parameter, when it expects a float64 here. If you return an int64 from the function, it will get turned into a bigint by clients. The JSON representation is `{ "$integer": "AQAAAAAAAAA=" }` not `1`.

PR #298, opened 2025-12-30
Add --typescript-compiler CLI flag for dev/deploy/run commands

smaccoun:
## Summary

Exposes the existing tsgo support via a CLI flag. This allows trying the experimental compiler without modifying convex.json.

### Why a CLI flag?

The main use case is CI/CD pipelines where you want to test tsgo compatibility without changing the committed config:

```yaml
# Test with tsgo in a separate CI job
- run: npx convex deploy --typescript-compiler=tsgo
```

This lets teams validate tsgo works for their project before committing to it in convex.json. Useful while tsgo is stabilizing, since you can run comparison builds or catch regressions without affecting the rest of the team.

### Changes

Added `--typescript-compiler` flag to:
- `convex dev`
- `convex deploy`
- `convex run --push`

### Usage

```bash
convex dev --typescript-compiler=tsgo
convex deploy --typescript-compiler=tsgo
```

Requires `@typescript/native-preview` to be installed when using tsgo.

### Resolution order

CLI flag takes precedence over convex.json, which takes precedence over the default (tsc). This follows the existing pattern for other options.

## Test plan

- [ ] Verify `convex dev --typescript-compiler=tsgo` uses tsgo when @typescript/native-preview is installed
- [ ] Verify `convex dev --typescript-compiler=tsc` uses tsc
- [ ] Verify default behavior (no flag) still works
- [ ] Verify `convex deploy --typescript-compiler=tsgo` works

### DIFF
--- a/npm-packages/convex/src/cli/deploy.ts
+++ b/npm-packages/convex/src/cli/deploy.ts
@@ -190,6 +190,7 @@ async function deployToNewPreviewDeployment(
     verbose?: boolean | undefined;
     typecheck: "enable" | "try" | "disable";
     typecheckComponents: boolean;
+    typescriptCompiler?: "tsc" | "tsgo" | undefined;
     codegen: "enable" | "disable";
 
     debug?: boolean | undefined;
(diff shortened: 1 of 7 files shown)


smaccoun:
On reflection, this is pretty minor. Closing.

PR #311, opened 2026-01-11
added --from-file flag for multi-line env vars

### LINKED ISSUE #128: convex env set fails to handle multi-line environment variable values
The `convex env set` command fails when trying to set environment variables that contain multi-line values, such as PEM-formatted private keys. This is particularly problematic for JWT authentication setups where `JWT_PRIVATE_KEY` needs to be a multi-line PEM-formatted RSA private key.

### Steps to Reproduce

1. Generate a standard PEM-formatted private key:
```bash
openssl genrsa -out private.key 2048
openssl pkcs8 -topk8 -inform PEM -outform PEM -nocrypt -in private.key -out private.pem
```

2. Try to set it as an environment variable:
```bash
convex env set JWT_PRIVATE_KEY "$(cat private.pem)" --admin-key <admin-key> --url <convex-url>
```

### Expected Behavior

The multi-line PEM key should be stored correctly as an environment variable.

### Actual Behavior

The command fails with various errors depending on the attempt:
- Direct attempt: Command hangs or fails
- With escaped newlines (`\n`): May appear to succeed but the value is stored incorrectly
- Via stdin redirect: Also fails

### Environment

- Convex CLI version: [latest]
- OS: Linux/macOS
- Context: Self-hosted Convex backend

### Impact

This blocks automated setup of JWT authentication in CI/CD pipelines for self-hosted Convex deployments (which i use for spinning up test envs). Manual intervention through the dashboard is required, which breaks automation workflows.

### Workarounds Attempted

1. Replacing newlines with `\n` - stores literal `\n` instead of actual newlines
2. Single line with spaces - breaks PEM format
3. Base64 encoding - requires application code changes to decode
4. Using stdin - still fails
5. Using heredoc - still fails

### Suggested Fix

The CLI should properly handle multi-line values, perhaps by:
- Supporting a `--from-file` flag: `convex env set JWT_PRIVATE_KEY --from-file private.pem`
- Properly parsing multi-line strings from command arguments
- Supporting stdin input: `cat private.pem | convex env set JWT_PRIVATE_KEY --from-stdin`

### Related

This issue is mentioned in the Convex auth setup documentation: https://labs.convex.dev/auth/setup/manual

Lasdw6:
Added a `--from-file` flag to `convex env set` to support setting multi-line environment variable values (e.g. PEM-formatted keys).

This avoids issues with passing multi-line secrets via shell arguments and preserves newlines correctly. Documentation is updated to include the new flag.

Fixes #128.

----

### DIFF
--- a/npm-packages/convex/src/cli/env.ts
+++ b/npm-packages/convex/src/cli/env.ts
@@ -24,16 +24,26 @@ const envSet = new Command("set")
   .description(
     "Set a variable: `npx convex env set NAME value`\n" +
       "Read from stdin: `echo 'value' | npx convex env set NAME`\n" +
+      "Read from a file: `npx convex env set NAME --from-file path/to/file`\n" +
       "If the variable already exists, its value is updated.\n\n" +
       "A single `NAME=value` argument is also supported.",
   )
+  .option("--from-file <path>", "Read value from file")
   .configureHelp({ showGlobalOptions: true })
   .allowExcessArguments(false)
   .action(async (originalName, originalValue, _options, cmd) => {
     const options = cmd.optsWithGlobals();
     const { ctx, deployment } = await selectEnvDeployment(options);
     await ensureHasConvexDependency(ctx, "env set");
-    await envSetInDeployment(ctx, deployment, originalName, originalValue);
+    const fromFile = typeof options.fromFile === "string" ? options.fromFile : undefined;
+    const envOptions = fromFile ? { fromFile } : undefined;
+    await envSetInDeployment(
+      ctx,
+      deployment,
+      originalName,
+      originalValue,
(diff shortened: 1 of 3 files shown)


Lasdw6:
Thanks for the clarification, that makes sense.

For context, I implemented --from-file based on the discussion in #128, where multiple people reported STDIN failing for multi-line values and a collaborator mentioned they’d take a PR adding --from-file. That’s what I was aiming to address here.

Appreciate the guidance on the preferred direction.

ianmacartney:
@Lasdw6 sorry for the miscommunication. I believe this should already be supported by doing:

```sh
cat ./file.txt | npx convex env set FOO
```

PR #314, opened 2026-01-13
fix(mcp): improve error message formatting in runOneoffQuery

### LINKED ISSUE #313: [MCP] runOneoffQuery returns "[object Object]" instead of actual error message
## Description
When `runOneoffQuery` fails, the error message shows `[object Object]` instead of the actual error details.

Example output:
```json
{"error":"Failed to run query: [object Object]"}
```

## Root Cause
In `npm-packages/convex/src/cli/lib/mcp/tools/runOneoffQuery.ts`, the catch block uses:

```typescript
catch (err) {
  ctx.crash(1, "fatal", `Failed to run query: ${(err as Error).toString().trim()}`);
}
```

When `err` is a plain object (not an Error instance), `.toString()` returns `[object Object]`.

## Suggested Fix

```typescript
catch (err) {
  let errorMessage: string;
  if (err instanceof Error) {
    errorMessage = err.message;
  } else if (typeof err === 'string') {
    errorMessage = err;
  } else {
    errorMessage = JSON.stringify(err, null, 2);
  }
  ctx.crash(1, "fatal", `Failed to run query: ${errorMessage}`);
}
```

## Impact
This affects debugging when using the Convex MCP tools - users cannot see what actually went wrong with their query.

dever4eg:
## Summary
- Fix error handling in `runOneoffQuery` to properly format non-Error objects
- When `err` is a plain object, `JSON.stringify` is used instead of `.toString()` which returns `[object Object]`

## Test plan
- [x] Trigger an error in `runOneoffQuery` (e.g., invalid query syntax)
- [x] Verify error message shows actual error details instead of `[object Object]`

Fixes #313

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/npm-packages/convex/src/cli/lib/mcp/tools/runOneoffQuery.ts
+++ b/npm-packages/convex/src/cli/lib/mcp/tools/runOneoffQuery.ts
@@ -104,10 +104,18 @@ export const RunOneoffQueryTool: ConvexTool<
         logLines: result.logLines,
       };
     } catch (err) {
+      let errorMessage: string;
+      if (err instanceof Error) {
+        errorMessage = err.message;
+      } else if (typeof err === "string") {
+        errorMessage = err;
+      } else {
+        errorMessage = JSON.stringify(err, null, 2);
+      }
       return await ctx.crash({
         exitCode: 1,
         errorType: "fatal",
-        printedMessage: `Failed to run query: ${(err as Error).toString().trim()}`,
+        printedMessage: `Failed to run query: ${errorMessage}`,
       });
     }
   },


mikecann:
Thanks, ill take a look at this very soon :)

mikecann (approved):
okay I have tested this and it fixed the error as expected, thanks!

Nicolapps:
Hi @dever4eg,

Thanks for the fix! Could you please include this message as part of the PR description?

> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

dever4eg:
> Hi @dever4eg,
> 
> Thanks for the fix! Could you please include this message as part of the PR description?
> 
> > By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Sure! Added

Nicolapps:
Thanks for your PR! We discovered that the root cause of the issue is a missing `toString` implementation in `RequestCrash`. We’re going to fix this issue instead, and close this PR since it doesn’t fix the root cause and could potentially be an issue if `err` is a non-JSON serializable value.

PR #323, opened 2026-01-22
Fix database reference in document-ids.mdx

yuto343:
Considering the context, it seems the table name is incorrect here.


Nicolapps (approved):
Thanks!

Nicolapps:
I forgot to give an update here, but your fix was merged in https://github.com/get-convex/convex-backend/commit/28ef9c90a5af. Thanks again!

Issue #287, opened 2025-12-22
Remove max 40 character limit for environment variables


danthegoodman1:
This is pretty easy to hit, for something like `WORKPOOL_IDENTIFICATION_REQUEST_MAX_PARALLELISM`

```
Failed to analyze case_materials/case_materials.js: Uncaught Error: The environment variable name WORKPOOL_IDENTIFICATION_REQUEST_MAX_PARALLELISM is too long. Environment variable names must be less than 40.
```

Is there a technical reason there's such a low limit of 40?

nipunn1313:
The main thing - is that the key->value mapping is entirely stored in memory and must be passed into javascript environment. For node actions, it has to happen on every request.

So we generally want to cap how much  total key+value space is used. I think it would probably be wiser to instead measure the total key+value size and have a net cap on that - instead of a smorgasbord of limits (name length 40, value length 8kb, max of 1000 env vars)

https://github.com/get-convex/convex-backend/blob/16315040105f25a9a66f895cfa750f800eb9c0c1/crates/common/src/types/environment_variables.rs#L66

nipunn1313:
I would be excited about holistically
- increasing the key length for environment variables by a lot (128?). It seems like some OSes have system limits at 256, and I don't want to be playing around near those limits.
- Setting a total limit on the total size of the entirety of keys/values
- Removing the # of env vars limit (currently 1000)

I'm a little hesitant about fiddling with one of these without handling the whole situation - and thinking about grandfathering.

I think a good gameplan would be

1) add a *new* limit on total size without enforcement. `tracing::warn` on startup if an existing deployment is already above the limit
2) Handle/communicate with everyone affected by (1)
3) Enforce limit from (1)
4) Remove env var limit + up the key size limit

We probably won't have time to prioritize this, but if someone from the community would take on 1,3,4 - I would help with 2.

yash-sangwan:
@nipunn1313 can I start working on this?

yash-sangwan:
This makes sense to me. I can break this into two PRs.

## PR 1

- Add a `MAX_TOTAL_ENV_VAR_SIZE = 1 MiB` constant
- Add a helper to calculate total env var size (names + values)
- Emit a `tracing::warn` when deployments exceed this total size
- Keep all existing limits as-is:
  - 40-character name limit
  - 8 KB value limit
  - 1000 env var count limit

This should give us real data on how many deployments are affected before enforcing anything.

## PR 2 (after communication)

[...]

PR #330, opened 2026-01-27
Max len env var name 40->128

okomarov:
Implements the soft suggestion from https://github.com/get-convex/convex-backend/issues/287 to increase the maximum length of an environment variable name from 40 to 128. 

Often, auto-generated variables, e.g. `NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY` exceed the 40-character limit. This avoids breaking the development flow or truncating/changing names which deviates from recommendations and/or documentation ( increasing future debugging time)

----

### DIFF
--- a/crates/common/src/types/environment_variables.rs
+++ b/crates/common/src/types/environment_variables.rs
@@ -63,7 +63,7 @@ static NAME_REGEX: LazyLock<Regex> =
 // don't reduce them since that might break existing projects.
 
 /// Maximum length of the name of an environment variable
-pub const MAX_NAME_LENGTH: usize = 40;
+pub const MAX_NAME_LENGTH: usize = 128;
 /// Maximum length of an environment variable value. 8KiB corresponds to the
 /// maximum length of an HTTP header.
 pub const MAX_VALUE_LENGTH: usize = 8 * (1 << 10);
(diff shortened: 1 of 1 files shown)


reneleonhardt:
Great to see a first fix!
Is the test regex correct, starting with `_`?

nipunn1313:
per the original comment - we can't simply make this change in a vacuum as it will break node actions.

But y'all are welcome to run a fork of convex with this PR to get long env vars.

If you want to send a PR to make it configurable by an env var in knobs.rs - would take that as well.

okomarov:
@nipunn1313 

> per the original comment - we can't simply make this change in a vacuum as it will break node actions.

This is not clear from your original comment. You've got a limit of 1000 vars with 8kb on value content. The max increase of the key name is 128-40 * 1000 vars = 88 kb. It's the allowance for 11 variables... 

> But y'all are welcome to run a fork of convex with this PR to get long env vars.
> 
> If you want to send a PR to make it configurable by an env var in knobs.rs - would take that as well.

We all have limited time and there is more time spent by everyone on an unreasonable limit of 40 characters than simply upping the limit by a bit. If 128 is too much, 64 would probably save a lot of time already.

If "It (Convex) makes it easy to build and scale dynamic live-updating apps." but shoots you in the foot when you're trying to integrate with another third party because you now need to figure out what's going on with env vars set by said third party, then...

I feel you've got more to gain from this simple PR but if I missed something else pls let me know

nipunn1313:
longer env var names make sense as a feature request! We're tracking it with #287.

It's on the radar, but need to do the multi step process to create a total-env-var size limit, since env vars are sent to node actions on every request. Until then, we'll keep the limit as is as simply increasing it will break other features.

Issue #345, opened 2026-02-06
Type instantiation error when file and folders have the same name


sindreij:
I've found a issue that makes typescript error with `Type instantiation is excessively deep and possibly infinite`. The issue happens when a file and a folder have the same name, so that `api.d.ts` ends up looking like (for instance).

```
declare const fullApi: ApiFromModules<{
  feed: typeof feed;
  "feed/items": typeof feed_items;
}>;
```

For some reason, the error only shows up when a query in `feed` returns a type that includes `any`, or `unknown`.

I have a minimal reproduction here: https://github.com/takex3/type-instantiation-minimal

(Note: I think I have a fix, that involves a small change to the `FilterApi` util type. However I create both an issue and a PR, since if the PR is rejected, the issue is still valid)

ianmacartney:
Landed in https://github.com/get-convex/convex-backend/commit/6b496fb939e764fa0fb0213feb5c51b4fba54084 ! Thanks!

PR #346, opened 2026-02-06
Fix infinite recursion when file and folders have same name

sindreij:
When a file and a folder has the same name, api.d.ts in the generated code ends up looking something like this:

```ts
declare const fullApi: ApiFromModules<{
  feed: typeof feed;
  "feed/items": typeof feed_items;
}>;
```

`ApiFromModulesAllowEmptyNodes` then generates this

```ts
{
  feed: {
    getFeedItem: FunctionReference<..>;
    items: {
      getFeedItemDisplayTitle: FunctionReference<..>,
    };
  };
}
```

When this is sent to `FilterApi` it creates a infinite recursion. I think the reason is that this part of FilterApi

```
API[mod] extends Predicate
    ? mod
    : API[mod] extends FunctionReference<any, any, any, any>
      ? never
      : FilterApi<API[mod], Predicate> extends Record<string, never>
        ? never
        : mod
```

ends up with a werirdness in the typescript-compiler where `mod` in `API[mod]` on line 5 ends up beeing not `"getFeedItem"` as we want, but `"getFeedItem" | "items"`. Full disclosure, I don't understand exactly how this works, and for getting to the issue I've had help by Codex, however 100% of the PR has been handcoded, the AI has only been used for researching.

Either way, the fix is pretty simple. We extract the part that returns never to a new util function. here the types end up behaving how we want them too

Fixes: #345

----

### DIFF
--- a/npm-packages/convex/src/server/api.ts
+++ b/npm-packages/convex/src/server/api.ts
@@ -270,6 +270,14 @@ type ApiFromModulesAllowEmptyNodes<AllModules extends Record<string, object>> =
     >
   >;
 
+type FilterKeysInApi<key, API, Predicate> = API extends Predicate
+  ? key
+  : API extends FunctionReference<any, any, any, any>
+    ? never
+    : FilterApi<API, Predicate> extends Record<string, never>
+      ? never
+      : key;
+
 /**
  * @public
  *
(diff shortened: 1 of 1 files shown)


ianmacartney:
Seems reasonable - can you get tests passing?

sindreij:
@ianmacartney: Tests are failing because 

> Credential access key has length 0, should be 32

So I'm guessing they won't run with the correct secrets since I'm not in your organization

ianmacartney:
I built a version of convex with this and it behaved as expected - working on upstreaming it

ianmacartney:
Landed in https://github.com/get-convex/convex-backend/commit/6b496fb939e764fa0fb0213feb5c51b4fba54084 ! Thanks!

georgecollier-nqu:
This has been driving me insane and made my codebase a mess with manual function definitions. Thank you for fix!

PR #348, opened 2026-02-07
Add v.object() unknownKeys: "strip" mode

robelest:
## Summary

Adds `unknownKeys` option to `v.object()` validators, controlling how extra fields are handled:
- `"strict"` (default): rejects documents with unknown fields (existing behavior)
- `"strip"`: silently removes unknown fields during validation

Stripping applies uniformly across all validation points:
- Document writes (insert/patch/replace/import)
- UDF argument validation
- UDF return value validation

## Key design decisions

**Union matching semantics**: strict-first, then strip by declaration order. When a document could match multiple union variants, strict validators are tried first. Among strip variants, the first declared match wins.

**Single-pass validate+strip**: `DocumentSchema::validate_and_strip` replaces the previous two-pass approach (check_new_document + strip_new_document), eliminating a redundant union walk on every write.

**Short-circuit when nothing to strip**: Both args and document paths check `has_extra_fields` before calling `strip_unknown_fields`, avoiding unnecessary allocations in the common case where the caller/writer sends exactly the declared fields.

**Pending schemas only validate, never strip**: Documents written during the brief pending→active schema transition window may retain extra fields.

## Backward compatibility

- Default mode is `Strict` everywhere — no behavior change for existing schemas
- JSON format omits `unknownKeys` field for Strict — old schemas parse correctly
- TS overloads preserve existing call signatures

## Design for future Passthrough

The `UnknownKeysMode` enum is designed so adding a `Passthrough` variant is straightforward:
- `strips_unknown_fields()` would return `false` for Passthrough
- `check_fields` skips extra-field rejection when `strips_unknown_fields()` is true
- JSON serialization handles arbitrary string values via `FromStr`
- TS type union just needs `| "passthrough"` added

---


npm-packages/convex/src/values/validators.ts
```diff
@@ -348,14 +365,21 @@ export class VObject<
    */
   omit<K extends keyof Fields & string>(
     ...fields: K[]
-  ): VObject<Expand<Omit<Type, K>>, Expand<Omit<Fields, K>>, IsOptional> {
+  ): VObject<
+    Expand<Omit<Type, K>>,
+    Expand<Omit<Fields, K>>,
+    IsOptional,
+    string,
```
ianmacartney: I think this loses the extracted field paths - maybe we can use the ObjectFieldPaths helper?

npm-packages/convex/src/values/validators.ts
```diff
@@ -368,7 +392,9 @@ export class VObject<
   ): VObject<
     Expand<Pick<Type, Extract<keyof Type, K>>>,
     Expand<Pick<Fields, K>>,
-    IsOptional
+    IsOptional,
+    string,
```
ianmacartney: ditto

npm-packages/convex/src/values/validators.ts
```diff
@@ -386,7 +413,9 @@ export class VObject<
   partial(): VObject<
     { [K in keyof Type]?: Type[K] },
     { [K in keyof Fields]: VOptional<Fields[K]> },
-    IsOptional
+    IsOptional,
+    string,
```
ianmacartney: ditto

npm-packages/convex/src/values/validators.ts
```diff
@@ -281,12 +281,8 @@ export class VObject<
...
   Fields extends Record<string, GenericValidator>,
   IsOptional extends OptionalProperty = "required",
-  FieldPaths extends string = {
-    [Property in keyof Fields]:
-      | JoinFieldPaths<Property & string, Fields[Property]["fieldPaths"]>
-      | Property;
-  }[keyof Fields] &
-    string,
+  FieldPaths extends string = ObjectFieldPaths<Fields>,
+  UnknownKeys extends "strict" | "strip" = "strict",
```
ianmacartney: until we have "passthrough" - I don't think we need this to be reflected in types, which would help with the auto-assigning of FieldPaths (not having to duplicate it or pass "string"). we can cross the "passthrough" bridge later (if ever)

npm-packages/system-udfs/convex/_system/server.ts
```diff
@@ -77,7 +77,8 @@ function withArgsValidated<T>(wrapper: T): T {
         if (!validateArgsResult.valid) {
           throw new Error(validateArgsResult.message);
         }
-        const functionResult = await functionDefinition.handler(ctx, args);
+        const validatedArgs = jsonToConvex(validateArgsResult.args[0]) as any;
+        const functionResult = await functionDefinition.handler(ctx, validatedArgs);
```
ianmacartney: was this a bug before? I don't think this is doing what you want- it'll turn a bigint into { $bigint: "AAAA..." }

npm-packages/convex/src/values/validator.ts
```diff
@@ -166,9 +168,25 @@ export const v = {
...
+      unknownKeys: options?.unknownKeys ?? "strict",
+    });
+  }) as {
+    <T extends PropertyValidators>(
+      fields: T,
+    ): VObject<ObjectType<T>, T>;
+    <T extends PropertyValidators, UK extends "strict" | "strip">(
+      fields: T,
+      options: { unknownKeys: UK },
+    ): VObject<ObjectType<T>, T, "required", ObjectFieldPaths<T>, UK>;
```
ianmacartney: hopefully dropping the UnknownKeys type will simplify this

ianmacartney (dismissed):
didn't go through all of it, but I think dropping the new type parameter will simplify a lot

ianmacartney (commented):
The API looks good. 

I want to see more usage of this in JS however - e.g. having some JS that defines these looser object validators as args / returns / schema and runs them "for real" - from Rust-based integration test runner or otherwise.

I'm going to add @Nicolapps to review the Rust side and provide guidance on testing approaches here.

It also makes me realize we should update convex-test to support this around the same time, so folks writing tests see the same behavior. Can you install a local build of convex in `convex-test` (you can use npm pack to make it) and put up a tentative PR there?

Also `convex-helpers` has `validate` and `parse` functions that should take this into account.

robelest:
Resolved merge conflicts with `main` and pushed `98489d561`.

Preserved the unknownKeys strip behavior and strict-first union matching. Ran targeted isolate/application/database strip-mode tests locally; all passed.

crates/application/src/application_function_runner/mod.rs
```diff
@@ -1466,13 +1466,20 @@ impl<RT: Runtime> ApplicationFunctionRunner<RT> {
...
-                {
-                    node_outcome.result = Err(js_err);
+                    ) {
+                        Err(js_err) => {
+                            node_outcome.result = Err(js_err);
+                        },
+                        Ok(validated_output) if validated_output.stripped => {
+                            node_outcome.result = Ok(validated_output.value);
+                        },
+                        Ok(_) => {},
```
dowski, line 1481: Is this expected to be impossible with the current API? Consider using `unreachable!()` if so.

crates/application/src/cache/mod.rs
```diff
@@ -621,13 +622,19 @@ impl<RT: Runtime> CacheManager<RT> {
...
-                                query_outcome.result = Err(js_err);
+                            ) {
+                                Err(js_err) => {
+                                    query_outcome.result = Err(js_err);
+                                },
+                                Ok(validated_output) if validated_output.stripped => {
+                                    query_outcome.result =
+                                        Ok(JsonPackedValue::pack(validated_output.value));
+                                },
+                                Ok(_) => {},
```
dowski, line 637: Ditto: re `unreachable!()`.

crates/common/src/schemas/json.rs
```diff
@@ -799,37 +822,31 @@ impl TryFrom<LiteralValidator> for JsonValue {
...
     fn try_from(o: ObjectValidator) -> anyhow::Result<BTreeMap<String, FieldTypeJson>> {
-        let mut map = BTreeMap::new();
-        for (field, field_type) in o.0 {
-            map.insert(field.to_string(), FieldTypeJson::try_from(field_type)?);
-        }
-        Ok(map)
+        o.fields
+            .into_iter()
+            .map(|(field, ft)| Ok((field.to_string(), FieldTypeJson::try_from(ft)?)))
+            .collect::<anyhow::Result<_>>()
```
dowski, line 849: Is this just a refactor/simplification?

crates/application/src/application_function_runner/mod.rs
Nicolapps: We want to avoid using `unreachable!` in production code because it will make the process panic. In shared services, this would be pretty bad (especially in multitenant services). In most cases we want to throw an internal error that we can observe from Sentry.

dowski: Got it, thanks!

crates/common/src/schemas/json.rs
robelest: Yeah, just a refactor. switched the for loop to an iterator chain, and o.0 became o.fields since ObjectValidator is now a named struct.

PR #353, opened 2026-02-11
Add sslmode=disable for non-SSL PostgreSQL connections

### LINKED ISSUE #354: Issue: Add sslmode=disable for non-SSL PostgreSQL connections
## Summary

When using `--do-not-require-ssl` to connect Convex to PostgreSQL with self-signed or untrusted certificates (e.g., CNPG in Kubernetes environments), the connection fails because no explicit `sslmode` parameter is set for the non-SSL case.

## Problem

The PostgreSQL driver in `crates/clusters/src/lib.rs` only handles the `require_ssl=true` case, setting `sslmode=require`. When `require_ssl=false`, no `sslmode` parameter is set, leaving the connection behavior undefined.

PostgreSQL's libpq client may default to `sslmode=prefer` in this case, which will attempt SSL negotiation and fail when presented with a self-signed certificate.

## Use Case

This is needed when:
- Running Convex self-hosted with CNPG (CloudNativePG) in Kubernetes/Coder
- Using PostgreSQL with self-signed certificates
- `DO_NOT_REQUIRE_SSL=true` environment variable is set

## Current Behavior

```rust
if require_ssl {
    cluster_url
        .query_pairs_mut()
        .append_pair("sslmode", "require");
} else {
    // Nothing here - sslmode is undefined!
}
```

## Expected Behavior

When `require_ssl=false`, explicitly set `sslmode=disable` to skip SSL verification:

```rust
if require_ssl {
    cluster_url
        .query_pairs_mut()
        .append_pair("sslmode", "require");
} else {
    cluster_url
        .query_pairs_mut()
        .remove_matching(|(key, _)| key == "sslmode") // Remove any existing sslmode
        .append_pair("sslmode", "disable");
}
```

## Additional Changes

Also update help text in `crates/local_backend/src/config.rs` to clarify that `--do-not-require-ssl` explicitly **disables** SSL, not just "doesn't require" it.

Current misleading text:
> It would still prefer SSL if available.

Should be:
> Disables SSL/TLS for PostgreSQL connections. When set, connects with sslmode=disable instead of sslmode=prefer.

## Workaround

No workaround exists - users cannot use self-hosted Convex with self-signed certificates without modifying the code.

jovermier:
Allows Convex to connect to PostgreSQL with self-signed certificates when `DO_NOT_REQUIRE_SSL=true` is set.

When `require_ssl` is false, the code now explicitly sets `sslmode=disable` instead of leaving it unset. This is needed for PostgreSQL with self-signed certificates (e.g., CNPG in Kubernetes environments).

Fixes #354 

----

### DIFF
--- a/crates/clusters/src/lib.rs
+++ b/crates/clusters/src/lib.rs
@@ -76,6 +76,10 @@ pub fn persistence_args_from_cluster_url(
                 cluster_url
                     .query_pairs_mut()
                     .append_pair("sslmode", "require");
+            } else {
+                cluster_url
+                    .query_pairs_mut()
+                    .append_pair("sslmode", "disable");
             }
             if require_leader {
                 cluster_url
--- a/crates/clusters/src/lib.rs
+++ b/crates/clusters/src/lib.rs
@@ -79,6 +79,9 @@ pub fn persistence_args_from_cluster_url(
             } else {
                 cluster_url
                     .query_pairs_mut()
+                    // Remove any existing sslmode before adding disable
+                    .query_pairs_mut()
+                    .remove_matching(|(key, _)| key == "sslmode")
                     .append_pair("sslmode", "disable");
             }
             if require_leader {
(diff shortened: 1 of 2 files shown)


goffrie (changes requested):
If you must disable SSL then it should suffice to add `&sslmode=disable` explicitly to your `POSTGRES_URL`.

PR #362, opened 2026-02-15
Clarify the licensing of backend and dashboard

typed-sigterm:
FSL is fair source, not open source. Because it doesn't comply with Article 6 of the OSI Open Source Definition **No Discrimination Against Fields of Endeavor**. Fair Source's official website also acknowledges this. Although the impact is not significant, it's still best to clarify.

### DIFF
--- a/npm-packages/docs/docs/home.mdx
+++ b/npm-packages/docs/docs/home.mdx
@@ -13,9 +13,10 @@ import { LargeCardList } from "@site/src/QuickstartsList.tsx";
 import { YouTubeList } from "@site/src/YouTubeLink.tsx";
 import Link from "@docusaurus/Link";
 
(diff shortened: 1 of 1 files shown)


nipunn1313:
Hi. Thanks for the PR.
We're aware of the distinction here - but are sticking with the "Open Source" phrasing - as it's the meaning in common parlance.

I do like some of the clarification on the purpose of FSL in your PR - so I'm going to pull in your change with some modifications - and I'll reference `fair.io` as it seems to have a good description of the distinction here.

nipunn1313:
https://github.com/get-convex/convex-backend/commit/9e6a8289f7c1ccee25c636b0d083b4e7dcdb6823

thx for contrib. I modified it a bit.

PR #368, opened 2026-02-20
Add tutorial async_hooks integration experiments and benchmarks

robelest:
## Summary
- add tutorial Convex actions that validate `node:async_hooks` / `AsyncLocalStorage` behavior with OpenTelemetry and LangChain/LangGraph integration paths
- add a focused benchmark action to estimate per-promise async_hooks overhead in the tutorial environment
- include tutorial dependency and lockfile updates needed to run these experiments locally

## Notes
- this PR is experiments-only and intentionally keeps runtime implementation changes in a separate PR
- `langgraph_test.invoke()` still surfaces an upstream `web-streams-polyfill` `_reader` collision and is kept as a known unrelated issue

### DIFF
--- a/npm-packages/demos/tutorial/convex/bench_test.ts
+++ b/npm-packages/demos/tutorial/convex/bench_test.ts
@@ -1,4 +1,53 @@
 import { action } from "./_generated/server";
+import { context as otelContext, trace } from "@opentelemetry/api";
+import { AsyncLocalStorageContextManager } from "@opentelemetry/context-async-hooks";
+
+class NoopSpanExporter {
+  export(
+    _spans: unknown[],
+    resultCallback: (result: { code: number }) => void,
+  ) {
+    resultCallback({ code: 0 });
+  }
+
+  shutdown() {
+    return Promise.resolve();
+  }
+
+  forceFlush() {
+    return Promise.resolve();
+  }
+}
+
+function errorMessage(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+function installPerformanceShimIfNeeded(): () => void {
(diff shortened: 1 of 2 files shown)


ianmacartney:
One thing I'd love to know is if we there's something that works with Vercel's AI SDK built-in telemetry - following their docs I ended up with dead ends with their axiom & otel integrations (too many other node deps), but maybe a thinner version of those libraries would work. @robelest

robelest:
Pushed latest telemetry experiment commit: `b86d70bfa`.

Ran local deploy + end-to-end mock telemetry validation:
- `just convex deploy -y --typecheck disable`
- `just convex run otel_test:testAiSdkTelemetry`
- `just convex run bench_test:benchPromises`

Results:
- `testAiSdkTelemetry`: all checks pass for generate, stream, `recordInputs:false`, `recordOutputs:false`, tool-call telemetry (`ai.toolCall`), concurrency isolation, error path, and telemetry-disabled path.
- `benchPromises`: `aiSdkGenerateText` benchmark now reports telemetry overhead. Latest run:
  - no telemetry median: `32ms`
  - with telemetry median: `100ms`
  - overhead vs no telemetry: `212.5%`

Notes:
- In this mock stream path, we consistently observe `ai.streamText.doStream`; top-level `ai.streamText` may be omitted.
- Tool-call flow completes and executes tool exactly once; mock flow can return empty final text while still emitting `ai.toolCall` telemetry.

ianmacartney:
These results are showing the AI SDK telemetry works but is slow - is that right? how does this compare to running in a node action (where there is a native AsyncLocalStorage)? Is our implementation much worse? Or is this just overhead associated with telemetry itself?

PR #369, opened 2026-02-21
feat(schema): add FlowFields, ComputedFields, and FlowFilters

ImRLopezAI:
## Summary

This PR proposes **schema-native derived fields** in Convex via three additions:

- **FlowFields**: read-time cross-table aggregations (`count`, `sum`, `avg`, `min`, `max`, `lookup`, `exist`)
- **ComputedFields**: row-level expressions derived from stored fields + FlowFields
- **FlowFilters**: typed runtime parameters for parameterized FlowField aggregations

The implementation adds end-to-end support across TypeScript schema APIs, Rust schema parsing/validation, runtime read-time resolution, codegen, and tests.

## Why This Helps Convex

Today, many teams solve derived values by:

- manual denormalization in mutations/triggers
- ad-hoc query-side recomputation
- component-level indirection for aggregates

This proposal makes derived data a **first-class schema feature**:

- fewer write-path responsibilities
- explicit and type-safe declaration in one place
- better consistency between schema intent and runtime behavior

### 1) Schema API additions (TS)

- `table.flowField(name, config)`
- `table.computed(name, config)`
- `table.flowFilter(name, config)`

`TableDefinition.export()` now serializes `flowFields`, `computedFields`, and `flowFilters`.

### 2) Type-system / DX improvements

- Generated **document read types** include FlowFields + ComputedFields.
- **Writable types** (`insert`, `patch`, `replace`) exclude read-only derived fields.
- FlowFilter references support typed `$field` autocomplete.

This improves developer ergonomics by preventing invalid writes at compile time while keeping reads expressive.

### 3) Backend runtime behavior (Rust)

- New resolver module: `crates/database/src/flow_fields.rs`
- On document reads/pages, derived fields are resolved and merged into returned objects.
- FlowFields read from source-table index ranges and are aggregated at read time.
- ComputedFields use a JSON expression DSL (`$add`, `$sub`, `$mul`, `$div`, `$gt/$gte/$lt/$lte/$eq/$ne`, `$cond`, `$concat`, `$ifNull`, field refs).

### 4) Reactivity integration

FlowField source index reads are recorded in the read set so updates to source rows invalidate subscribers appropriately.

### 5) Validation + tests

Coverage spans:

- TS schema/type-level tests (`npm-packages/convex/src/server/schema.test.ts`)
- Rust schema parsing/validation tests (`crates/common/src/schemas/tests.rs`)
- Expression/aggregation unit tests (`crates/database/src/flow_fields.rs`)

## Comparison with Aggregate Component

Related existing solution:

- [Aggregate component](https://www.convex.dev/components/aggregate)

This proposal is **complementary**, not a replacement.

Where FlowFields are stronger:

- schema-native declaration in core Convex (no extra component wiring)
- per-document derived values returned directly with normal reads
- parameterized aggregations via FlowFilters
- integrated codegen + write-safety (derived fields excluded from write types)
- simpler mental model for common derived-field use cases

Where Aggregate component remains strong:

- explicit denormalized aggregate pipelines
- high-throughput/global counters and aggregate-heavy workloads
- cases where component boundaries are preferred for architecture reasons

### Why core integration matters

Because FlowFields are resolved in the core read path, developers can avoid many app-level aggregation patterns that rely on large `.collect()`-style reads. In practice, this can reduce exposure to large payload/read-limit failure modes (for example, 16MB+ read errors) in common aggregation scenarios.

## Developer Experience Improvements

- Single source of truth in `schema.ts` for derived behavior
- Fewer manual update paths in mutations
- Safer writes (derived fields are read-only at type level)
- Better readability and discoverability of business logic

## Demo / Reproduction

I built a full demo app here:

- [ImRLopezAI/flow-field](https://github.com/ImRLopezAI/flow-field)

The demo showcases:

- `customers` table with `orderCount`, `totalSpent`, `balanceLCY` FlowFields
- `tier` and `displayName` ComputedFields
- runtime `dateFilter` and `dimensionFilter` FlowFilters
- reactive UI updates when source data changes

## Testing

- Local proposal and demo validation completed.
- Demo tests pass (author-reported 100%).

## Notes / Scope

- This proposal is **open to API changes** based on maintainer feedback.
- It is also open to additional follow-up features (for example: batching optimizations, caching strategy, and expanded expression operators).
- Goal of this PR: start technical review on API shape, runtime semantics, and integration strategy.


ianmacartney:
Am I understanding correctly that this could mean reading one document
could result in an ~unbounded amount of other documents getting read
automatically, e.g. if there was a count?

ImRLopezAI:
Hi Ian,

Thats why a relatianal field is required to avoid this complexity you can
check the show case, and at the end these 3 type of "fields" will be read
only and they wont be stored on the database

https://github.com/ImRLopezAI/flow-field/blob/42c58a77e2803e9c2920e8e4137e82008e430197/convex/schema.ts#L8-L74

ianmacartney:
In that example, it seems reading one customer document could end up loading every related document. If that customer had a million leger entries, reading that single document would cause the DB to need to scan & sum a million documents - which is against the Convex design principle of making things explicit and limiting reads in transactions. It would also mean reading a customer in a transaction would conflict with any other transaction writing to the orders or leger for the customer. 
Unfortunately I don't think this is a workable direction for the product. Adding something like this would likely need to do the work on the write side to keep reads cheap, keeping materialized views, which is what we recommend folks do themselves via triggers or otherwise

PR #371, opened 2026-02-22
Simplify useQuery object form to minimal QueryOptions base

robelest:
## Summary
- add a minimal object-form `useQuery` overload that accepts `QueryOptions<Query> | "skip"`
- keep return type aligned with legacy `useQuery`: `T | undefined`, and keep throwing query errors
- remove object-form extras from this first step (`initialValue`, `throwOnError`, and preloaded object support)
- rename `ConvexQueryOptions` to `QueryOptions` and keep `convexQueryOptions()` helper typed to that base shape
- keep `usePreloadedQuery` as a thin wrapper by factoring preloaded parsing into `parsePreloaded`

## Why
This keeps phase 1 intentionally small and avoids locking in richer result/error semantics before we settle the final shape.

## Testing
- `npm run format-check` (in `npm-packages/convex`)
- `npm run test-not-silent -- src/react/use_query.test.ts src/react/use_query_object_options.test.ts` (in `npm-packages/convex`)

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #372, opened 2026-02-23
Return object-form useQuery result state and add throwOnError

robelest:
## Summary
- follow up on #371 by changing object-form `useQuery` to return `UseQueryResult<T>` (`data` / `error` / `status`)
- add `throwOnError?: boolean` for object-form `useQuery` (default `false`), while keeping legacy positional `useQuery(query, args)` behavior unchanged
- keep `error` in the result type regardless of `throwOnError` to avoid overload complexity and keep typing predictable
- export `UseQueryResult` from `convex/react`
- add runtime and type tests for object-form return typing + `throwOnError`

## Notes
- stacked on #371, so this PR includes the base commit until #371 merges.

## Testing
- `npm run format-check` (in `npm-packages/convex`)
- `npm run test-not-silent -- src/react/use_query.test.ts src/react/use_query_object_options.test.ts src/react/use_query_result.test.ts src/react/client.test.tsx` (in `npm-packages/convex`)

----

### DIFF
--- a/npm-packages/convex/src/react/client.ts
+++ b/npm-packages/convex/src/react/client.ts
@@ -802,6 +802,33 @@ export type OptionalRestArgsOrSkip<FuncRef extends FunctionReference<any>> =
     : [args: FuncRef["_args"] | "skip"];
 
 /**
+ * Result returned by object-form {@link useQuery}.
+ *
+ * @public
+ */
+export type UseQueryResult<QueryResult> =
+  | {
+      data: QueryResult;
+      error: undefined;
+      status: "success";
+    }
+  | {
+      data: undefined;
+      error: Error;
+      status: "error";
+    }
+  | {
+      data: undefined;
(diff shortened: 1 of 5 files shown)


npm-packages/convex/src/react/client.ts
```diff
@@ -801,66 +801,113 @@ export type OptionalRestArgsOrSkip<FuncRef extends FunctionReference<any>> =
...
+/**
+ * Result returned by object-form {@link useQuery}.
+ *
+ * @public
+ */
+export type UseQueryResult<QueryResult> =
+  | {
+      data: QueryResult;
+      error: undefined;
+      isPending: false;
```
ianmacartney: snip isPending

robelest:
Moved to internal repo: https://github.com/get-convex/convex/pull/47801 (squashed with #371)

PR #373, opened 2026-02-23
Add object-form options overload to usePaginatedQuery

robelest:
## Summary
- add object-form overloads for `usePaginatedQuery` and `usePaginatedQuery_experimental`
- support calling with `{ query, args, initialNumItems }` while keeping legacy positional overloads unchanged
- keep existing return shape and statuses unchanged (`results`, `isLoading`, `status`, `loadMore`)
- add coverage for object-form `args: \"skip\"` and initial loading behavior in `use_paginated_query.test.tsx`

## Why
This aligns pagination with the object-form query direction while keeping the API minimal and backwards compatible.

## Notes
- stacked on #372 (which is stacked on #371), so this PR includes prior commits until they merge.

## Testing
- `bun run test-not-silent -- src/react/use_query.test.ts src/react/client.test.tsx src/react/use_paginated_query.test.tsx` (in `npm-packages/convex`)

----

### DIFF
--- a/npm-packages/convex/src/react/use_paginated_query.test.tsx
+++ b/npm-packages/convex/src/react/use_paginated_query.test.tsx
@@ -22,6 +22,9 @@ import type {
   PaginatedQueryArgs,
   PaginatedQueryItem,
   PaginatedQueryReference,
+  UsePaginatedQueryObjectErrorResult,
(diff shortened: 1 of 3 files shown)


robelest:
Moved to internal repo: https://github.com/get-convex/convex/pull/47810

PR #375, opened 2026-02-23
Fix upgrading.md

xixixao:
The instructions no longer make any sense after https://github.com/get-convex/convex-backend/commit/9f917ca08de8675bf2a526a25bacb0eae9238d27

----


xixixao:
@nipunn1313 those instructions don't make sense though either, since migrations were moved to `crates/migrations_model` ? But I'm also not clear on which releases I need from the git log?

PR #377, opened 2026-02-25
docs: update convex-test docs for Vitest v4

luchillo17:
Updates the convex-test documentation:
- Replace environmentMatchGlobs with projects array (Vitest v4 approach)
- Remove server.deps.inline from config examples (Vitest handles deps automatically)

### DIFF
--- a/npm-packages/docs/docs/testing/convex-test.mdx
+++ b/npm-packages/docs/docs/testing/convex-test.mdx
@@ -62,8 +62,7 @@ convex-test library.
   <Step title="Configure Vitest">
 
     Add <JSDialectFileName name="vitest.config.ts" /> file to configure the test
-    environment to better match the Convex runtime, and to inline the test library
-    for better dependency tracking.
+    environment to better match the Convex runtime.
 
     <Details summary={<>If your Convex functions are in a directory other than <code>convex</code></>}>
       If your project has a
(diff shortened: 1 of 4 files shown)


ianmacartney:
Looks reasonable - I can't recall why we had it inlining convex-test - that's the biggest open question.
@Nicolapps can you review & upstream this if it looks good to you?

npm-packages/docs/docs/testing/convex-test.mdx
```diff
@@ -104,22 +103,34 @@ convex-test library.
...
       import { defineConfig } from "vitest/config";
 
       export default defineConfig({
         test: {
-          environmentMatchGlobs: [
-            // all tests in convex/ will run in edge-runtime
-            ["convex/**", "edge-runtime"],
-            // all other tests use jsdom
-            ["**", "jsdom"],
+          projects: [
```
Nicolapps: @cursor Update this to include both the Vitest 3 and Vitest 4 instructions (Vitest 4 first). In both cases, don't include `server: { deps: { inline: ["convex-test"] } },`

xixixao:
> Specifies modules that should be transformed and resolved by Vite. These modules are run by Vite's module runner.

> Typically, your source files are inlined.

I'd wager this might have eased bundling, otherwise maybe we'd need to do the whole esm/cjs rigmarole in convex-test like we do in convex. The other thing I remember caring about was Vitest detecting changes (in tests and project sources) and rerunning the tests. If both work, I'd say go for it.

Nicolapps:
@luchillo17 Thank you for your contribution! I merged it in 6b72618c3edd44afac4c61052d7facc409c06fb7

I also restored guidance for Vitest 3 in 3c5a711f7933e85207bc14be6b69c1306b6d0644 and tests for the `projects` guidance in 1119f05

PR #384, opened 2026-02-27
dashboard: add copy buttons to deployment URLs in summary

jl33-ai:
## Summary
- Adds inline copy buttons next to the **Cloud URL** and **HTTP Actions URL** in the deployment summary/overview panel
- Uses the existing `CopyButton` component (with `inline` mode and tooltips) for a consistent UX
- The URLs remain clickable links — the copy button appears alongside them

## Motivation
The deployment overview page displays the Cloud URL and HTTP Actions URL as clickable links, but there's no easy way to copy them to the clipboard without manually selecting the text. The settings page already has copy buttons for these URLs via `CopyTextButton`, but the overview/summary panel (which is the first thing users see) does not. This is a small quality-of-life improvement.

## Test plan
- [ ] Verify the copy button appears next to both Cloud URL and HTTP Actions URL in the deployment summary
- [ ] Verify clicking the copy button copies the URL to clipboard and shows "Copied!" feedback
- [ ] Verify the URL link remains clickable and opens in a new tab
- [ ] Verify layout doesn't break on smaller viewports (the flex container handles wrapping)

### DIFF
--- a/npm-packages/dashboard-common/src/features/health/components/DeploymentSummary.tsx
+++ b/npm-packages/dashboard-common/src/features/health/components/DeploymentSummary.tsx
@@ -18,6 +18,7 @@ import {
   CodeIcon,
   ExternalLinkIcon,
 } from "@radix-ui/react-icons";
+import { CopyButton } from "@common/elements/CopyButton";
 import Link from "next/link";
 import { useContext, useEffect, useState } from "react";
 import semver from "semver";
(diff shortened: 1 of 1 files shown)


atrakh:
Hi @jl33-ai, thank you for the suggestion and contribution. 

We won't be accepting this at this time as most browsers natively support a quick-copy of links with the right-click context menu

jl33-ai:
what?
p-1-report-jony-ive-checked-out-at-apple-years-ago

PR #382, opened 2026-02-27
Add object-form client override for query hooks

robelest:
## Summary
- add optional `client?: ConvexReactClient` to object-form query options used by `useQuery` and `useSuspenseQuery`
- thread the optional client through query parsing and pass it into `useQueries`
- extend `useQueries` to accept an optional client override and fall back to `useConvex()` when omitted
- keep suspense behavior aligned with #374:
  - same skip sentinel semantics
  - suspense options do not accept `throwOnError`
- add runtime and type tests covering object-form client override behavior

## Why
This enables per-query client routing for object-form hooks without changing legacy positional behavior.

## Notes
- stacked on #374 (which is stacked on #373/#372/#371), so this PR includes prior commits until those merge.

## Testing
- `bun run format-check` (in `npm-packages/convex`)
- `bun run test-not-silent -- src/react/use_query.test.ts src/react/use_query_object_options.test.ts src/react/use_suspense_query.test.ts src/react/use_queries.test.ts src/react/client.test.tsx` (in `npm-packages/convex`)

----
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #374, opened 2026-02-23
Add useSuspenseQuery as a thin wrapper with skip support

robelest:
## Summary
- add `useSuspenseQuery` with positional and object overloads
- implement it as a thin wrapper around `useQuery`, sharing object-result semantics from #372
- keep a single skip sentinel behavior across hooks:
  - active query suspends until data arrives
  - skipped query does not suspend and returns `undefined`
- keep suspense error handling boundary-first (do not expose `throwOnError` on suspense options)
- export `useSuspenseQuery` from `convex/react`
- add runtime and type tests for success, suspense loading, skip behavior, and suspense option typing

## Notes
- stacked on #373 (which is stacked on #372/#371), so this PR includes prior commits until they merge.

## Testing
- `bun run format-check` (in `npm-packages/convex`)
- `bun run test-not-silent -- src/react/use_query.test.ts src/react/use_query_object_options.test.ts src/react/use_suspense_query.test.ts src/react/client.test.tsx` (in `npm-packages/convex`)

----
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #385, opened 2026-02-28
Add requireAuth option for object-form query hooks

robelest:
## Summary
- add `requireAuth?: boolean` to object-form `useQuery` and `useSuspenseQuery` options
- gate subscriptions behind auth state when `requireAuth` is true, returning pending/no-suspend while auth is loading or unauthenticated
- throw a usage error when `requireAuth: true` is used without a Convex auth provider context
- extract shared auth context plumbing into an internal module to avoid a `client.ts`/`ConvexAuthState.tsx` import cycle
- add runtime and type tests for `requireAuth` behavior and option typing

## Why
This keeps auth-sensitive query flows explicit at the callsite, while preserving current positional behavior and avoiding unnecessary subscriptions before auth is ready.

## Notes
- stacked on #382 (which is stacked on #374/#373/#372/#371), so this PR includes prior commits until those merge.

## Testing
- `bun run format-check` (in `npm-packages/convex`)
- `bun run test-not-silent -- src/react/ConvexAuthState.test.tsx src/react/client.test.tsx src/react/use_query.test.ts src/react/use_query_object_options.test.ts src/react/use_suspense_query.test.ts src/react/use_queries.test.ts` (in `npm-packages/convex`)
- `just rush build -t convex`

----

### DIFF
--- a/npm-packages/convex/src/react/ConvexAuthContext.ts
+++ b/npm-packages/convex/src/react/ConvexAuthContext.ts
@@ -1,0 +1,22 @@
+import { createContext, useContext } from "react";
+
+/**
+ * Type representing the state of an auth integration with Convex.
+ *
+ * @public
+ */
+export type ConvexAuthState = {
+  isLoading: boolean;
+  isAuthenticated: boolean;
+};
+
+export const ConvexAuthContext = createContext<ConvexAuthState | undefined>(
+  undefined,
+);
+
+/**
+ * @internal
+ */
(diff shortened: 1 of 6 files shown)


robelest:
Moved to internal repo. PR will be created shortly.

robelest:
Moved to get-convex/convex: https://github.com/get-convex/convex/pull/47992

robelest:
Moved to get-convex/convex: https://github.com/get-convex/convex/pull/47992

PR #392, opened 2026-03-07
fix: prefer native WebSocket over ws package for Bun compatibility

### LINKED ISSUE #390: Bun WebSocket client rejects successful 101 Switching Protocols during convex deploy against self-hosted Convex (works with Node)
## Summary

When deploying to a self-hosted Convex backend with Bun, `convex deploy` fails during websocket sync with:

- `WebSocket error message: Unexpected server response: 101`
- `WebSocket closed with code 1006`

This is surprising because HTTP `101 Switching Protocols` indicates a successful websocket upgrade.

## Environment

- Convex CLI: `convex@1.31.6`
- Bun: `1.3.10`
- OS: Linux (GitLab CI runner)
- Self-hosted Convex behind nginx + TLS
- Deploy command:
  - `bunx --bun convex@1.31.6 deploy --yes`
- Works with Node runtime:
  - `pnpm dlx convex@1.31.6 deploy --yes`

## Expected behavior

`convex deploy` should connect to the sync websocket and complete deployment successfully when server returns `101 Switching Protocols`.

## Actual behavior

Deployment starts, but websocket sync repeatedly fails and reconnects:

```text
- Deploying to https://backend-api.tischtenniscup.de...
WebSocket error message: Unexpected server response: 101
WebSocket closed with code 1006
Attempting reconnect in 1312ms
...
```

## What I validated already

- `CONVEX_SELF_HOSTED_URL` and admin key are correct.
- URL health check succeeds:
  - `GET /version` returns OK.
- Proxy and TLS are working.
- Websocket upgrade endpoint responds correctly (`101`).
- Backend logs show Bun reaches sync endpoint and receives `101`, then disconnects immediately.
- Development deployment works in CI, production fails with Bun.

## Relevant backend log pattern

Backend shows successful upgrade followed by immediate disconnect from Bun client:

```text
GET /api/1.31.6/sync HTTP/1.1" 101 ... "Bun/1.3.10"
WARN ... WebSocket protocol error: Connection reset without closing handshake: Client disconnected
```

## Reproduction command

```bash
bunx --bun convex@1.31.6 deploy --yes
```

## Notes

- Same Convex deployment works with Node-based invocation (`pnpm dlx convex@1.31.6 deploy --yes`).
- This seems isolated to Bun runtime path used by `convex deploy` websocket sync.

## Question

Is Bun officially supported for `convex deploy` against self-hosted deployments?  
If yes, is this a known issue/workaround?  
If no, please clarify recommended runtime for CI deploys.

bxff:
Fixes #390

The CLI and the Node.js `ConvexClient` always import the `ws` package and use it as the WebSocket constructor. This is unnecessary on runtimes that already have a native `WebSocket` (Bun, Deno, Node 22+), and it actively breaks things under Bun because Bun's `ws` shim mishandles HTTP 101 upgrades through reverse proxies.

This is technically a Bun bug, but preferring the native `WebSocket` when it exists is a good change on its own. The build script already externalizes `ws` for Bun compat (build.cjs L155), and `BaseConvexClient` already falls back to the global `WebSocket` when no constructor is passed (client.ts L293). This just extends that same pattern to the CLI entry points.

Can't drop `ws` entirely because we still support Node 18, which doesn't have a global `WebSocket`.

**What changed:**

New file `src/cli/lib/nodeWebSocket.ts` exports a `nodeWebSocket` that checks for a global `WebSocket` first, falls back to `ws`. Then `run.ts` and `networkTest.ts` import that instead of `ws` directly. `simple_client-node.ts` inlines the same check since it's bundled separately by esbuild.

**Repro:**

Set up a WebSocket server behind a Node HTTP proxy (simulating nginx) and connected from Bun 1.3.5. `ws` through the proxy times out, while direct connections work fine. Node handles both cases. Full verification with real nginx + TLS would need the reporter's setup.

### DIFF
--- a/npm-packages/convex/src/browser/simple_client-node.ts
+++ b/npm-packages/convex/src/browser/simple_client-node.ts
@@ -3,10 +3,15 @@ import {
   setDefaultWebSocketConstructor,
 } from "./simple_client.js";
 
+// Prefer the native global WebSocket when available (Bun, Deno, Node >= 21)
+// and fall back to the `ws` npm package for older Node.js versions.
 // This file is compiled with `bundle: true` with an exception for
-// `./simple_client.js` so this "ws" import will be inlined.
+// `./simple_client.js` so a dynamic import approach won't work here.
 import ws from "ws";
-const nodeWebSocket = ws as unknown as typeof WebSocket;
+const nodeWebSocket: typeof WebSocket =
+  typeof WebSocket !== "undefined"
+    ? WebSocket
+    : (ws as unknown as typeof WebSocket);
 
 setDefaultWebSocketConstructor(nodeWebSocket);
 
--- a/npm-packages/convex/src/cli/lib/networkTest.ts
+++ b/npm-packages/convex/src/cli/lib/networkTest.ts
@@ -16,7 +16,7 @@ import {
   formatSize,
   ThrowingFetchError,
 } from "./utils/utils.js";
-import ws from "ws";
+import { nodeWebSocket } from "./nodeWebSocket.js";
 import { BaseConvexClient } from "../../browser/index.js";
 import { DefaultLogger } from "../../browser/logging.js";
 const ipFamilyNumbers = { ipv4: 4, ipv6: 6, auto: 0 } as const;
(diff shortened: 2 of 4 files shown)


bxff:
Probably not the correct fix.

PR #409, opened 2026-03-16
fix: accept camelCase `tableName` in streaming export endpoints

### LINKED ISSUE #408: `tableName` parameter silently ignored on `/api/document_deltas` and `/api/list_snapshot` — serde deserialization falls through to `Everything`
## Bug

The `tableName` query parameter on the streaming export endpoints (`/api/document_deltas` and `/api/list_snapshot`) is silently ignored. All tables are returned regardless of the `tableName` value. This affects both the Airbyte and any HTTP client following the [official API docs](https://docs.convex.dev/streaming-export-api), which document the parameter as `tableName` (camelCase).

## Root cause

In [`crates/fivetran_source/src/api_types/mod.rs`](https://github.com/get-convex/convex-backend/blob/main/crates/fivetran_source/src/api_types/mod.rs), the `SelectionArg` enum is:

```rust
#[derive(Serialize, Deserialize)]
#[serde(untagged)]
pub enum SelectionArg {
    Exact { selection: Selection },
    SingleTable {
        table_name: String,       // expects snake_case "table_name"
        component: Option<String>,
    },
    SingleComponent { component: String },
    Everything {},                // catch-all fallback
}
```

The parent structs `DocumentDeltasArgs` and `ListSnapshotArgs` have `#[serde(rename_all = "camelCase")]`, but this does **not propagate** to the flattened `#[serde(untagged)]` enum. So:

1. The [API docs](https://docs.convex.dev/streaming-export-api) say the parameter is `tableName` (camelCase)
2. The [Airbyte source-convex connector](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-convex/source_convex/source.py#L127) sends `tableName` (camelCase)
3. The server expects `table_name` (snake_case)
4. Because the enum is `#[serde(untagged)]`, serde tries each variant in order. `tableName` doesn't match `table_name`, so `SingleTable` fails silently. It falls through to `Everything {}`, which returns **all tables unfiltered**.

The existing tests confirm snake_case is expected:

```rust
fn test_table_name_only() {
    assert_selection_arg_deserialization(
        r#"{"table_name": "users"}"#,  // snake_case
        SelectionArg::SingleTable { ... },
    );
}
```

## Impact

Every Airbyte user syncing from Convex is affected. Each configured stream receives **all tables** instead of just its own, causing:

- **N-fold data duplication** — with 22 streams configured, the entire database is synced 22 times (e.g. 57M records instead of 2.6M)
- **Massive sync slowdown** — a sync that should take ~30 min takes 9-14 hours
- **Storage waste in the destination** — every destination table contains all Convex record types, not just the expected one

This has existed since the Airbyte connector was introduced (Nov 2022).

## Suggested fix

Add `#[serde(alias = "tableName")]` to accept both formats:

```rust
SingleTable {
    #[serde(alias = "tableName")]
    table_name: String,
    component: Option<String>,
},
```

Or add `#[serde(rename_all = "camelCase")]` to the `SelectionArg` enum itself.

A corresponding test should be added:

```rust
#[test]
fn test_table_name_camel_case() {
    assert_selection_arg_deserialization(
        r#"{"tableName": "users"}"#,
        SelectionArg::SingleTable {
            table_name: "users".to_string(),
            component: None,
        },
    );
}
```

## Environment

- Convex deployment version: 1.31.7
- Airbyte source-convex connector (latest from registry)
- Bug exists on `main` as of 2025-09-03 (commit `75dd02d2c4`)

tom-eberle:
## Summary

- Adds `#[serde(alias = "tableName")]` to `SelectionArg::SingleTable` so the streaming export endpoints (`/api/document_deltas` and `/api/list_snapshot`) accept the camelCase `tableName` parameter documented in the API docs and used by the Airbyte connector
- Adds tests for camelCase deserialization (`tableName` alone and with `component`)

Fixes #408

## Details

The `SelectionArg` enum uses `#[serde(untagged)]`, so when `tableName` (camelCase) doesn't match the expected `table_name` (snake_case), serde silently falls through to `Everything {}`, returning all tables unfiltered. The parent structs have `#[serde(rename_all = "camelCase")]` but this doesn't propagate into flattened untagged enums.

## Test plan

- [ ] Existing `test_table_name_only` still passes (snake_case still works)
- [ ] New `test_table_name_camel_case` passes
- [ ] New `test_table_name_camel_case_and_component` passes
- [ ] `cargo test -p fivetran_source`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/fivetran_source/src/api_types/mod.rs
+++ b/crates/fivetran_source/src/api_types/mod.rs
@@ -139,6 +139,7 @@ pub enum SelectionArg {
 
     /// If only the table name is provided, assumes it’s in the root component.
     SingleTable {
+        #[serde(alias = "tableName")]
         table_name: String,
 
         /// The component path of the table. If not provided, the table is
@@ -226,6 +227,28 @@ mod tests {
             r#"{"selection": { "_other": "incl" }}"#,
             SelectionArg::Exact {
                 selection: Selection::default(),
+            },
+        );
+    }
+
+    #[test]
+    fn test_table_name_camel_case() {
+        assert_selection_arg_deserialization(
+            r#"{"tableName": "users"}"#,
+            SelectionArg::SingleTable {
+                table_name: "users".to_string(),
+                component: None,
+            },
+        );
+    }
+
+    #[test]
+    fn test_table_name_camel_case_and_component() {
+        assert_selection_arg_deserialization(
+            r#"{"tableName": "jobs", "component": "cron"}"#,
+            SelectionArg::SingleTable {
+                table_name: "jobs".to_string(),
+                component: Some("cron".to_string()),
             },
         );
     }


Nicolapps:
Merged in f38b02c65dbfbb11afc018ff15611d2d9a8c37a1

PR #412, opened 2026-03-19
fix(posthog): detect invalid project tokens via /decide endpoint

richardsolomou:
## Problem

PostHog's ingestion endpoints (`/i/v1/logs` and `/i/v0/e/`) return HTTP 200 regardless of project token validity — they're fire-and-forget by design. This meant `verify_creds()` always succeeded, even with garbage tokens, and the integration never entered the "Failed" state.

## Changes

- Switch both `PostHogLogsSink` and `PostHogErrorTrackingSink` to validate project tokens against PostHog's `/decide?v=3` endpoint, which returns 401 for invalid tokens
- Remove the now-dead `is_verification` parameter from `send_batch` since `verify_creds` no longer calls it
- Fix decide URL construction to preserve port for self-hosted instances by cloning the full URL instead of using `host_str()`
- Rename "Project API Key" to "Project Token" in dashboard forms and docs to match PostHog's own nomenclature (Settings > Project > General)
- Update tests to register a `/decide` handler and remove verification event filtering

## How did you test this code?

- All 6 existing PostHog tests pass (bad token, normal operation, bandwidth tracking for both sinks)
- Verified against PostHog's `/decide` endpoint source code that it returns 401 for invalid tokens

----

### DIFF
--- a/crates/local_backend/src/log_sinks.rs
+++ b/crates/local_backend/src/log_sinks.rs
@@ -525,7 +525,7 @@ fn validate_posthog_host(host: Option<&String>) -> anyhow::Result<()> {
 #[derive(Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct CreatePostHogLogsLogStreamArgs {
-    /// PostHog project API key.
+    /// PostHog project token.
     api_key: String,
     /// PostHog host URL. Defaults to https://us.i.posthog.com.
     host: Option<String>,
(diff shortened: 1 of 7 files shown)


nipunn1313:
bugbot run

nipunn1313:
bugbot run

nipunn1313:
bugbot run

nipunn1313:
cool - I'm just going to merge this as is - even though there are a couple issues. Seems good enough now.

nipunn1313:
landed in https://github.com/get-convex/convex-backend/commit/9f0eee5d00aae8edb7b712e234f58cba1b066e51

PR #415, opened 2026-03-25
fix: use self-hosted dashboard URL in auth config error message

### LINKED ISSUE #56: using clerk with self-hosting gives incorrect dashboard link from CLI
The `npx convex dev` command on the CLI gives a link to this incorrect dashboard page with `null` as the deployment name when working with self-hosting.

Can repro by setting up convex self-hosting and setting up clerk.
Ideally the dashboard link here should link to the localhost:6791 self-hosted dashboard.

```
npx convex dev
Saved the given --url as CONVEX_URL to .env.local
✖ Environment variable CLERK_JWT_ISSUER_DOMAIN is used in auth config file but its value was not set. Go to:

    https://dashboard.convex.dev/d/null/settings/environment-variables?var=CLERK_JWT_ISSUER_DOMAIN

  to set it up. 
```

majiayu000:
Fixes #56

## Summary

When `handlePushConfigError()` encounters an `AuthConfigMissingEnvironmentVariable` error for self-hosted deployments, it falls through to a generic message without a dashboard link. This is because `deploymentName` is `null` for self-hosted setups (no `deploymentFields`).

This adds an `else if` branch that detects self-hosted mode via the `CONVEX_SELF_HOSTED_URL` env var and constructs the dashboard URL using `hostname:6791` (the default self-hosted dashboard port from `docker-compose.yml`) with the path `/settings/environment-variables?var=VAR_NAME`.

## Test plan

- All existing `config.test.ts` tests pass (14/14)
- Manual verification: set `CONVEX_SELF_HOSTED_URL=http://localhost:3210` and trigger the `AuthConfigMissingEnvironmentVariable` error path — output shows `http://localhost:6791/settings/environment-variables?var=VAR_NAME` instead of `dashboard.convex.dev`

### DIFF
--- a/npm-packages/convex/src/cli/lib/config.ts
+++ b/npm-packages/convex/src/cli/lib/config.ts
@@ -30,6 +30,7 @@ import {
   logAndHandleFetchError,
   ThrowingFetchError,
   currentPackageHomepage,
+  CONVEX_SELF_HOSTED_URL_VAR_NAME,
 } from "./utils/utils.js";
 import { recursivelyDelete } from "./fsUtils.js";
 import { NodeDependency } from "./deployApi/modules.js";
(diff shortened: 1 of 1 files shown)


ianmacartney:
Do you prefer the dashboard over `npx convex env set` ? Another way we could improve this would be to make the `env set` string include the name of the variable to set.
My only hesitation here is the hard-coding of the port, since you likely wouldn't have that port exposed for anything except when using localhost.

majiayu000:
Thanks for the feedback @ianmacartney\! You're right on both points — the hard-coded port 6791 would only be valid for localhost setups and isn't reliably exposed otherwise. I've updated the fix to instead include the variable name directly in the `npx convex env set` suggestion (e.g. `npx convex env set AUTH_SECRET <value>`), which works for all deployment types including self-hosted. The dashboard link for cloud deployments (when `deploymentName` is set) is unchanged.

PR #419, opened 2026-03-27
fix: pass database spec before storage flags in run_backend.sh

hahagu:
When both POSTGRES_URL (or MYSQL_URL) and S3 storage environment variables are set, `$DB_SPEC` is appended as a trailing positional argument after --s3-storage, causing the CLI to interpret the database connection string as the S3 storage argument: `--db postgres-v5 --s3-storage postgresql://user:pass@host:5432`

This causes the backend to attempt S3 operations against the Postgres URL, leading to high CPU usage, memory growth, and degraded performance over time.

The fix moves `$DB_SPEC` to immediately follow `${DB_FLAGS[@]}` so the database connection string is correctly associated with --db: `--db postgres-v5 postgresql://user:pass@host:5432 --s3-storage`

----


hahagu:
-

hahagu:
Found out that `./convex-local-backend` reads each as separate positional arguments.
Closing.

PR #439, opened 2026-04-09
feat: preserve schema field order in dashboard data browser

### LINKED ISSUE #383: Dashboard: column order should respect schema field order
## Summary

When viewing table data in the Convex dashboard, columns are not displayed in the order fields are defined in the `schema.ts` file. The current ordering appears to be alphabetical or based on when fields were first encountered, rather than following the `defineTable()` field definition order.

## Expected Behavior

Columns in the dashboard data view should default to the order in which fields are defined in the schema. For example, given:

```ts
export default defineSchema({
  users: defineTable({
    name: v.string(),
    email: v.string(),
    role: v.string(),
    createdAt: v.number(),
  }),
});
```

The dashboard should display columns as: `name | email | role | createdAt` (plus system fields like `_id`, `_creationTime`).

## Current Behavior

Columns appear in a different order that doesn't match the schema definition. While drag-and-drop reordering is available as a workaround, it:

- Doesn't sync across team members
- May not persist across browser storage clears
- Requires manual effort for every table

## Why This Matters

- Schema field order is intentional — developers group related fields and order them by importance
- Modern JS engines preserve object key insertion order, so `defineTable()` field order is deterministic
- Having the dashboard reflect the schema makes it easier to scan and verify data at a glance

## Suggestion

When a schema is defined, use the schema field order as the default column order in the dashboard data view. The existing drag-and-drop reordering could still override this default per user.

mvanhorn:
## Summary

When a table has enforced schema validation with complete fields, the data browser now preserves the schema-defined field order instead of sorting alphabetically.

## Why this matters

`sortColumns()` in `useTableFields.ts` always sorted fields alphabetically (`_id` first, `_creationTime` last). This destroyed the intentional ordering in the schema definition. As @xixixao noted in #383, Convex runtime objects are already sorted by key, so having the dashboard also ignore schema order makes it harder to reason about field layout.

`topLevelFieldsForValidator()` returns fields via `Object.keys(validator.value)` which preserves insertion order. The fix removes the `sortColumns()` call on the schema-validated path so that order is kept.

## Changes

- `useTableFields.ts`: Return `result.fields` directly (instead of `sortColumns(result.fields)`) when schema validation is enforced and fields are complete. Non-schema fallback paths still sort alphabetically.
- `useTableFields.test.tsx`: Added test verifying fields defined as `zebra, apple, middle` in the schema appear in that order, not alphabetically.

## Testing

Added test case "preserves schema field order instead of sorting alphabetically" to `useTableFields.test.tsx`. The test creates a schema with fields in non-alphabetical order and verifies the hook returns them in schema order.

Manual column reordering via drag-and-drop is unaffected - `DataContent.tsx` stores user-customized `columnOrder` in localStorage, which overrides the default.

Fixes #383

This contribution was developed with AI assistance (Codex).

### DIFF
--- a/npm-packages/dashboard-common/src/features/data/components/Table/utils/useTableFields.test.tsx
+++ b/npm-packages/dashboard-common/src/features/data/components/Table/utils/useTableFields.test.tsx
@@ -203,13 +203,14 @@ describe("useTableFields", () => {
       useTableFields("test", shape, activeSchema, []),
     );
 
-    // Fields should be in schema order (zebra, apple, middle) not alphabetical (apple, middle, zebra)
+    // Fields should be in schema order (zebra, apple, middle) not alphabetical,
+    // with _id first and _creationTime last
     expect(result.current).toEqual([
       "_id",
-      "_creationTime",
       "zebra",
       "apple",
       "middle",
+      "_creationTime",
     ]);
   });
 
--- a/npm-packages/dashboard-common/src/features/data/components/Table/utils/useTableFields.test.tsx
+++ b/npm-packages/dashboard-common/src/features/data/components/Table/utils/useTableFields.test.tsx
@@ -162,6 +162,57 @@ describe("useTableFields", () => {
     expect(result.current).toEqual(["_id", "name", "_creationTime"]);
   });
 
+  it("preserves schema field order instead of sorting alphabetically", () => {
+    const shape: Shape = {
+      type: "Record",
+      keyShape: { type: "String" },
+      valueShape: {
+        optional: false,
+        shape: { type: "Boolean" },
+      },
+    };
+    const activeSchema: SchemaJson = {
+      schemaValidation: true,
+      tables: [
+        {
+          tableName: "test",
+          documentType: {
+            type: "object",
+            value: {
+              zebra: {
+                fieldType: { type: "string" },
+                optional: false,
+              },
+              apple: {
+                fieldType: { type: "string" },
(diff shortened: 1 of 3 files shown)


npm-packages/dashboard-common/src/features/data/components/Table/utils/useTableFields.test.tsx
```diff
@@ -0,0 +1,249 @@
...
+    };
+
+    const { result } = renderHook(() =>
+      useTableFields("test", shape, activeSchema, []),
+    );
+
+    // Fields should be in schema order (zebra, apple, middle) not alphabetical (apple, middle, zebra)
+    expect(result.current).toEqual([
+      "_id",
+      "_creationTime",
```
ianmacartney: Currently _creationTime goes last I believe - can we maintain that by passing a param to `sortColumns` like "maintainOrder" that returns 0 if neither field is _id / _creationTime?

mvanhorn:
Good call - added a `maintainOrder` option to `sortColumns` in 304afd1. When true, it returns 0 for non-system fields so `_id` stays first and `_creationTime` goes last while preserving the schema-defined order for everything else.

ianmacartney:
landed in f01ea5f1f78c8a8fbd612144e99601203ad618af

mvanhorn:
Thanks for landing the schema field order preservation.

ianmacartney:
thanks for the PR!

PR #442, opened 2026-04-12
Fix `convex dev --start` exiting before cleanup finishes on Ctrl+C

### LINKED ISSUE #441: `convex dev --start` leaves `next dev --turbopack` running after `Ctrl+C`
## Summary

When running `convex dev` with `--start` through Bun's script runner, pressing `Ctrl+C` stops the Convex dev process but does not always stop the command started via `--start`.

In my case, the package script is:

```json
{
  "scripts": {
    "dev": "convex dev --start 'next dev --turbopack'"
  }
}
```

Running `bun run dev` can leave the Next.js dev server running in the background after `Ctrl+C`, which leaves orphaned processes around and can cause memory leaks or stale dev servers on the same port.

## Steps To Reproduce

1. Add a package script like:

   ```json
   {
     "scripts": {
       "dev": "convex dev --start 'next dev --turbopack'"
     }
   }
   ```

2. Run:

   ```sh
   bun run dev
   ```

3. Wait for both Convex and Next.js to start.
4. Press `Ctrl+C`.

## Expected Behavior

Both processes should shut down cleanly:

- `convex dev` should exit
- the `--start` command and its child processes should also exit

## Actual Behavior

- `convex dev` exits
- `next dev --turbopack` can continue running in the background

https://github.com/user-attachments/assets/241b487b-64ec-4a54-a0c9-f6fc3985cc2e

## Impact

- orphaned dev server processes
- memory usage continues growing after Convex exits
- port conflicts on the next run
- confusing local dev behavior because Convex is stopped but Next is still alive

## Environment

- OS: macOS
- Invocation: `bun run dev`
- Script: `convex dev --start 'next dev --turbopack'`
- Framework: Next.js dev server with Turbopack

## Notes

- After further testing, I can only reproduce this when Convex is invoked via Bun's script runner.
- Running Convex directly, for example:

  ```sh
  bunx convex dev --start 'next dev --turbopack'
  ```

  shuts down correctly on `Ctrl+C` in my setup.
- So this appears to be specifically about the `bun run` invocation path rather than `convex dev --start` being broken in all invocation modes.
- Here are some verbose logs from the failing `bun run` path: https://gist.github.com/brandon-julio-t/4b276c6bd8e3fb04ce502a51297ab3b5

Maybe relevant issues:
- https://github.com/npm/cli/issues/5021
- https://github.com/npm/cli/issues/8164
- https://github.com/npm/cli/issues/7693

brandon-julio-t:
## Summary
- forward `SIGINT` immediately to detached `--start` commands so sidecar shutdown begins as soon as `convex dev` receives `Ctrl+C`
- keep the cleanup-time `SIGTERM` fallback for `--start` commands that do not exit promptly on `SIGINT`
- use the standard `130` exit code for `convex dev` and `convex init`, and ignore Bun's immediate duplicate `SIGINT` while preserving a later `Ctrl+C` force-exit path

Closes #441.

## Testing
- `pnpm exec eslint src/cli/lib/dev.ts src/cli/dev.ts src/cli/init.ts`
- `node common/scripts/install-run-rush.js build -t convex`
- `bun run dev` from `farmio-kokopilot` with `"dev": "convex dev --start 'next dev --turbopack'"`
- press `Ctrl+C` and confirm both Convex and Next.js stop cleanly
- `bunx convex dev --start 'next dev --turbopack'` still shuts down correctly

https://github.com/user-attachments/assets/b8749dee-009e-4ecd-9b5f-357601764f5a

## Disclosure
- This change was prepared with assistance from GPT-5.4-xhigh and behaviors reviewed and verified by me.

### DIFF
--- a/npm-packages/convex/src/cli/dev.ts
+++ b/npm-packages/convex/src/cli/dev.ts
@@ -179,9 +179,25 @@ Same format as .env.local or .env files, and overrides them.`,
   .showHelpAfterError()
   .action(async (cmdOptions) => {
     const ctx = await oneoffContext(cmdOptions);
+    const duplicateSigintGraceMs = 500;
+    let cleanupStartTime: number | null = null;
     process.on("SIGINT", async () => {
+      if (cleanupStartTime !== null) {
+        // `bun run <script>` can deliver an immediate duplicate SIGINT while the
+        // first handler is starting cleanup. Ignore that short-window signal, but
+        // keep a later Ctrl+C as a force-exit escape hatch.
+        if (Date.now() - cleanupStartTime < duplicateSigintGraceMs) {
+          logVerbose(
+            "Received SIGINT during cleanup, ignoring duplicate signal...",
+          );
+          return;
+        }
+        logVerbose("Received SIGINT during cleanup, exiting immediately...");
+        process.exit(130);
+      }
+      cleanupStartTime = Date.now();
       logVerbose("Received SIGINT, cleaning up...");
-      await ctx.flushAndExit(-2);
+      await ctx.flushAndExit(130);
     });
 
     if (cmdOptions.deployment !== undefined) {
--- a/npm-packages/convex/src/cli/init.ts
+++ b/npm-packages/convex/src/cli/init.ts
@@ -21,9 +21,25 @@ export const init = new Command("init")
       adminKey: undefined,
       envFile: undefined,
     });
+    const duplicateSigintGraceMs = 500;
+    let cleanupStartTime: number | null = null;
     process.on("SIGINT", async () => {
+      if (cleanupStartTime !== null) {
+        // `bun run <script>` can deliver an immediate duplicate SIGINT while the
+        // first handler is starting cleanup. Ignore that short-window signal, but
+        // keep a later Ctrl+C as a force-exit escape hatch.
+        if (Date.now() - cleanupStartTime < duplicateSigintGraceMs) {
+          logVerbose(
+            "Received SIGINT during cleanup, ignoring duplicate signal...",
+          );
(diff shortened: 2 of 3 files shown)


ianmacartney:
Thanks for the PR! I don't love special-casing bun workarounds, but I agree the general behavior can be improved w.r.t. forwarding SIGINT. 
I think there may be a race where on shutdown it'll send SIGINT then almost immediately SIGTERM? But I'm looking at whether to factor this in a shared setup and give some graceful shutdown grace - waiting a bit for the process to exit before sending SIGTERM

brandon-julio-t:
> I think there may be a race where on shutdown it'll send SIGINT then almost immediately SIGTERM?

yes, I do think this is the root cause.

somehow npm and bun run will send SIGINT 2 times in quick succession, and 2 SIGINTs mean we kill the app, hence the cleanup was never run.

so to simulate:

**ideal scenario:**

[00:00:00]: SIGINT by ctrl+c
[00:00:00]: convex cleanup start
[00:00:01]: convex cleanup done
[00:00:02]: exits cleanup

**real scenario:**

[00:00:00]: SIGINT by ctrl+c
[00:00:00]: convex cleanup start
[00:00:01]: SIGINT by ctrl+c (mysterious, is a known issue even in npm)
[00:00:01]: exits immediately without finishing the cleanup

btw what CONVEX_VERBOSE sees are 2 SIGINTs, no SIGTERM.

ianmacartney:
This should have the fix in it:
```
npm i https://pkg.pr.new/get-convex/convex@da63701
```
Landed in 129a02a58f8be9a57fdcf9263525dc93f704cacf

Issue #446, opened 2026-04-17
Feature Request: Add generic invocation metadata for one-shot and nested Convex function calls


Mat4m0:
Convex already distinguishes between validated function arguments and runtime
execution context (`request_id`, `execution_id`, caller ancestry, scheduled job
ancestry).

This issue proposes extending that existing runtime context with optional,
generic invocation metadata that is:

- passed beside args, not inside args
- readable from `ctx.meta`
- propagated through nested `ctx.runQuery`, `ctx.runMutation`, and
  `ctx.runAction` calls

## Motivation

Today, if a framework or advanced integration wants to propagate request-scoped
metadata such as:

- correlation IDs
- tracing context
- origin hints
- workflow phase labels
- platform request context

there is no first-class place to put it.

The practical workaround is to include reserved metadata fields in function
args. That has a few downsides:

- function validators must accept framework internals
- business args and runtime context get mixed together
- metadata can leak into user code unless wrappers strip it out
- nested propagation has to be reimplemented outside the runtime

Convex already has the right abstraction boundary for this problem in
`ExecutionContext`. This would extend that boundary without changing the
function programming model.

## Proposal

Add optional invocation metadata to function calls and surface it through
`ctx.meta`.

Conceptually:

- one-shot callers accept optional metadata beside args
- nested `ctx.runQuery`, `ctx.runMutation`, and `ctx.runAction` accept optional
  metadata beside args
- `ctx.meta.getInvocationContext()` returns the current invocation context
  including metadata

Example client call:

```ts
await client.action(
  api.orders.process,
  { orderId: "order_123" },
  {
    metadata: {
      correlationId: "req_123",
      origin: "checkout",
    },
  },
);
```

Example server usage:

```ts
export const processOrder = action({
  args: { orderId: v.string() },
  handler: async (ctx, args) => {
    const invocation = await ctx.meta.getInvocationContext();

    return await ctx.runQuery(
      internal.orders.load,
      { orderId: args.orderId },
      {
        metadata: {
          ...invocation.metadata,
          phase: "authorize",
        },
      },
    );
  },
});
```

## Proposed behavior

- Metadata is optional.
- Metadata is passed beside args, not inside args.
- Nested calls inherit parent metadata by default.
- A nested call may pass `metadata` to shallowly override top-level keys.
- Metadata is available through `ctx.meta.getInvocationContext()`.
- Metadata is not auth context and should not be used in place of `ctx.auth`.

## Scope

[...]

PR #447, opened 2026-04-17
feat: Add generic invocation metadata to function calls and ctx.meta

Mat4m0:
Solves: #446 

## Summary

This PR adds optional invocation metadata to Convex function calls.

Invocation metadata is:

- passed beside args, not inside args
- propagated through nested `runQuery` / `runMutation` / `runAction`
- readable from `ctx.meta.getInvocationContext()`

This is an additive runtime primitive for request-scoped metadata such as
correlation IDs, tracing context, origin hints, and workflow phase labels.

### Runtime / backend

- Extended `ExecutionContext` with optional invocation metadata.
- Added protobuf round-trip support.
- Threaded metadata through public function execution paths.
- Threaded metadata through nested function execution paths.
- Added `1.0/getInvocationContext` to the async syscall surface.

### TypeScript API

- Added `InvocationMetadata`
- Added `InvocationOptions`
- Added `InvocationContext`
- Added `ctx.meta.getInvocationContext()`
- Updated nested `ctx.runQuery`, `ctx.runMutation`, and `ctx.runAction`
  signatures to accept optional invocation metadata
- Updated `ConvexHttpClient` query / mutation / action methods to accept
  optional invocation metadata

### Node executor / action callbacks

- Propagates inherited invocation metadata in callback headers
- Sends nested override metadata in callback bodies
- Decodes invocation metadata into `ExecutionContext` on callback ingress

### Docs / examples

- Added a new docs page for invocation metadata
- Added snippets for client-side and server-side usage
- Added cross-links from related docs

## Behavior

- Metadata is optional.
- Metadata is not part of `args`.
- Nested calls inherit parent metadata by default.
- Nested call metadata shallowly overrides top-level keys.
- Metadata is accessible via `ctx.meta.getInvocationContext()`.
- Reactive subscriptions are not in scope for this PR.

## Motivation

Without this primitive, frameworks and advanced integrations have to encode
request-scoped metadata into function args using reserved fields. That mixes
business input with runtime context and forces validators and wrappers to
account for framework internals.

This PR keeps business args pure and uses the existing execution-context layer
for runtime metadata.

## Non-goals

This PR does not add:

- subscription metadata semantics
- tracing APIs
- logging APIs
- auth changes

## Tests

Added targeted coverage for:

- metadata merge semantics
- execution context protobuf round-trip
- public metadata parsing
- callback header decoding
- HTTP client request shapes
- node-executor propagation
- `ctx.meta.getInvocationContext()` syscall behavior

## Notes

The implementation is generic and not framework-specific. It is intended as a
low-level runtime primitive that can support multiple higher-level use cases
without changing the Convex function model.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,8 @@
+{
+  "permissions": {
+    "allow": [
+      "Skill(pr-review-toolkit:review-pr)",
+      "Skill(pr-review-toolkit:review-pr:*)"
+    ]
+  }
+}
--- a/crates/application/src/api.rs
+++ b/crates/application/src/api.rs
@@ -57,6 +57,7 @@ use udf::{
 };
 use value::{
     sha256::Sha256Digest,
+    ConvexObject,
     DeveloperDocumentId,
 };
 
@@ -102,6 +103,7 @@ pub trait ApplicationApi: Send + Sync {
         path: ExportPath,
         args: SerializedArgs,
         caller: FunctionCaller,
+        invocation_metadata: Option<ConvexObject>,
         ts: ExecuteQueryTimestamp,
         journal: Option<SerializedQueryJournal>,
     ) -> anyhow::Result<RedactedQueryReturn>;
@@ -117,6 +119,7 @@ pub trait ApplicationApi: Send + Sync {
         path: CanonicalizedComponentFunctionPath,
         args: SerializedArgs,
         caller: FunctionCaller,
+        invocation_metadata: Option<ConvexObject>,
         ts: ExecuteQueryTimestamp,
         journal: Option<SerializedQueryJournal>,
     ) -> anyhow::Result<RedactedQueryReturn>;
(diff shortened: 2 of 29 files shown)


ianmacartney:
You've read my mind!
Coincidentally I have PRs up to add a bunch of metadata:
```ts
    const { deploymentName, deploymentType } = await ctx.meta.getDeploymentMetadata();
    const { componentPath, name, visibility, type } = await ctx.meta.getFunctionMetadata();
    const { bytesRead, documentsRead, ... } = await ctx.meta.getTransactionMetrics();
    
    // only in mutations & actions:
    const { ip, userAgent, requestId } = await ctx.meta.getRequestMetadata();
```
And we're also thinking about how to enrich logging so that you can more easily add tracing info. 

I am nervous about adding global metadata to queries that is accessible at runtime, as this makes caching queries ~impossible to do correctly, and instead would prefer for this data to be injected at the logging layer, so it can be uniformly injected on cache hits without re-running the UDF

Mat4m0:
Thanks for the fast answer :) this was helpful.

I can see the caching problem with request-scoped metadata in query runtime, and after auditing my own use case I think that concern is real.

Given that, I don’t think the current proposal is the right shape to keep pushing. The more promising direction seems to be:

- logging-layer enrichment across all lanes
- richer `ctx.meta.*` where cache-safe
- request-scoped metadata primarily in mutation/action lanes, not query runtime
- `ctx.setLogAttributes(...)` also looks especially relevant from my side.

Closing this for now

PR #457, opened 2026-04-26
Add custom test mutations

ezyyeah:
## What changed

This adds support for running custom test mutations alongside the existing custom test query flow.

It includes:

- a new `Custom mutation` action in the dashboard Data view
- backend support for running ad hoc mutations through `run_test_function`
- a new deploy key permission for custom test mutations
- CLI support for one off inline mutations
- MCP support for one off mutations
- docs and story updates for the new flow

## Notes

I noticed that the existing custom test query flow uses a fixed REPL style module path.

With custom test mutations, that behavior becomes more important because the temporary module path should not risk overwriting or deleting a real module entry during cleanup.

To address that, I split the branch into two commits:

1. `feat: add custom test mutation runners`
2. `fix: generate fresh repl module paths for test runners`

The second commit updates the temp REPL module path handling so each run gets a fresh server generated path.

I wanted to call this out explicitly because the fixed REPL path behavior was already present for custom test queries. If this should behave differently, or if there is a preferred pattern for temporary REPL modules, I would appreciate feedback.

## Testing

Tested with:

- local dashboard flow for `Custom mutation`
- local CLI flow for inline mutations
- targeted Rust tests for REPL temp module path generation
- `cargo build -p local_backend --bin convex-local-backend`

----


ianmacartney:
Thanks for this! We've discussed it before. It's a pretty big risk / has a lot of gotchas to think through. I'll stew on it more, but off the top of my head:

1. If you start a mutation that uses the scheduler (e.g. to do recursive pagination), how do you stop it if you accidentally made an infinite loop / fork bomb. 
2. If you start another mutation and the previous one was still scheduling itself, is the new code a replacement for the same UDF
3. Can you import code, e.g. to use your own custom functions - right now the inline query can't import your own code. This would be a big win for folks who want their aggregates to stay up to date, e.g.
4. What are the related permissions here - and making sure admins with audit logging requirements can control this vector.

For 1/2, an easy thing here would be to either not support the scheduler and/or not allow that function to be scheduled / executed - e.g. ctx.meta.getFunctionMetadata wouldn't work, or would give a name that is non-routable somehow, but this feels a bit hacky.

PR #461, opened 2026-04-30
feat(common): make environment variable name max length configurable

officialasishkumar:
### Summary

The environment variable name length limit has been hard-coded at 40 since the original implementation. Users keep hitting it because third-party SDKs ship with auto-generated identifiers that exceed 40 characters, e.g. `NEXT_PUBLIC_SUPABASE_PUBLISHABLE_DEFAULT_KEY` (44), `WORKPOOL_IDENTIFICATION_REQUEST_MAX_PARALLELISM` (47), `TRIGGER_OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` (46), `ANTHROPIC_SCORE_PHONE_TEXT_MESSAGES_DOCUMENT_MODEL` (50). When the SDK reads such a variable at module load time, `convex deploy` aborts with `The environment variable name ... is too long. Environment variable names must be less than 40.`, which blocks the integration entirely.

[#287](https://github.com/get-convex/convex-backend/issues/287) collects several reports of this. As a follow-up to that thread, [@nipunn1313 explicitly invited a PR](https://github.com/get-convex/convex-backend/issues/287#issuecomment-3677470014) that makes the limit configurable via `knobs.rs`, so self-hosted operators can opt into longer names today, even before the broader rework around a total-size budget lands. This PR is that change.

### Changes

- Add `ENV_VAR_NAME_MAX_LENGTH` knob in `crates/common/src/knobs.rs`, defaulting to `40` so cloud and existing self-hosted deployments see no behavior change.
- Switch `EnvVarName::from_str` to read the knob (and surface the configured value in the error message).
- Drop the now-unused `MAX_NAME_LENGTH` constant. Nothing else in the workspace referenced it.

### Behavior

- **Default:** unchanged. The limit is still `40`, so existing projects, the cloud product, and CI keep behaving exactly as before.
- **Self-hosted opt-in:** operators can set `ENV_VAR_NAME_MAX_LENGTH=128` (or any other value) in the backend's environment to accept longer names.
- The total budget is still bounded by the existing `ENV_VAR_LIMIT` (count) and `ENV_VAR_TOTAL_SIZE_LIMIT` (bytes) knobs, so this does not regress the per-request payload size sent to node actions.

### Validation

- `cargo check -p common`, `cargo check -p model`, `cargo check -p application`, `cargo check -p local_backend` all succeed on `nightly-2026-02-18`.
- `rustfmt --check` is clean for the changed files.
- The validator continues to reject the existing failure modes (invalid characters, leading digit), and a 41-character name is still rejected with the default knob.

Refs #287

----

### DIFF
--- a/crates/value/src/base32.rs
+++ b/crates/value/src/base32.rs
@@ -62,6 +62,33 @@ pub const fn encoded_buffer_len(len: usize) -> usize {
 /// 3 bytes past the last chunk & that `data[len..]` is all zeroes
 #[inline]
 pub fn encode_into<const MAY_OVERREAD: bool>(out: &mut [u8], data: &[u8], len: usize) {
+    cfg_select! {
+        all(
+            target_arch = "aarch64",
+            target_endian = "little",
+            target_feature = "neon"
+        ) => {
+            // SAFETY: this block is gated behind `target_feature = "neon"`.
+            unsafe {
+                encode_into_neon::<MAY_OVERREAD>(out, data, len)
+            }
+        }
+        _ => {
+            encode_into_inner::<MAY_OVERREAD>(out, data, len)
+        }
+    }
+}
+
+#[inline]
+#[cfg(any(
+    test,
+    not(all(
+        target_arch = "aarch64",
+        target_endian = "little",
+        target_feature = "neon"
+    ))
+))]
+fn encode_into_inner<const MAY_OVERREAD: bool>(out: &mut [u8], data: &[u8], len: usize) {
     // Process the input in chunks of length 5 (i.e 40 bits).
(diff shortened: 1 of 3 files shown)


ianmacartney:
Thanks! I have already started making some changes to up the default name length to 256, with an overall size limit. I'm assuming that's sufficient for you, so I think we can avoid adding extra knobs.

ianmacartney:
Now that the env var name can be up to 256 characters, I think this is not necessary - agreed @officialasishkumar ?

officialasishkumar:
Agreed, now that main allows env var names up to 256 characters with the total size limit in place, the configurable name-length knob is no longer necessary. I rebased the PR branch onto main and dropped the obsolete change.

PR #464, opened 2026-05-05
Fix dry-run push to roll back schema changes and skip index creation

zack-stoke:
# Fix dry-run push to roll back schema changes and skip index creation

## What's broken

`convex deploy --dry-run` is not dry. Since `dcc7519ce` ("Use components push path as default", Oct 2025), every dry-run push commits schema changes to the real deployment and leaves them there. Concretely:

- A new schema is written to `_schemas` in `Pending`/`Validated` state
- That schema is enforced on live writes immediately — if the dry-run schema is more restrictive than the current active schema, production writes start failing
- The schema stays enforced until the next real deploy overwrites it
- New indexes are created and the IndexWorker starts backfilling them

The only thing dry-run correctly skips is function activation in `finish_push`.

## How it got here

The history across three commits tells the story:

**September 2024** (`33e4a3eee`) — dry-run was enabled for the components push path for the first time, removing the `"dryRun not allowed yet"` crash. The design assumed `start_push` was non-mutating for dry runs. It wasn't — `StartPushRequest` had no `dry_run` field on the backend, so the flag the CLI sent was silently dropped and schema changes were committed unconditionally.

**January 2025** (`04a69fdfe`) — `_handle_schema_change_in_start_push` was introduced with an explicit `if dry_run { drop(tx) }` branch. This was correct: the transaction was dropped without committing for dry runs, so no schema or index changes were persisted.

**October 2025** (`dcc7519ce`) — the `dry_run` branch and the `dry_run` field from `StartPushRequest` were removed, with the comment: *"Even in dry run mode, we need to commit the schema changes so that wait_for_schema can validate the schema against existing data."* The intent — schema validation against real data — is sound. The gap: `finish_push` was updated to skip function activation for dry runs, but nothing was added to roll back the schema changes that were now unconditionally committed. The comment *"We can't actually run finish_push in a dry run, since we rolled back all of our changes during start push"* described intended behavior that was never implemented.

## What this PR does

**1. Wire `dry_run` through the push pipeline**

Adds `dry_run: bool` (with `#[serde(default)]` for backwards compatibility) to `StartPushRequest` and `WaitForSchemaRequest`. Introduces a `PushMode` enum (`Normal` / `DryRun`) that propagates through the stack as a typed value rather than a raw boolean.

**2. Roll back schema changes in `finish_push`**

When `dry_run: true`, `finish_push` now calls `rollback_dry_run_schema_change` before returning. This calls `mark_overwritten` on each schema document created by `start_push`, which:
- Transitions the schema `Pending`/`Validated` → `Overwritten`
- Stops `enforce` from validating live writes against the dry-run schema
- Deletes the `_schema_validation_progress` document, signalling the SchemaWorker to stop

**3. Skip index creation entirely for dry runs**

In `start_component_schema_changes`, dry runs now call `get_index_diff` instead of `prepare_new_and_mutated_indexes`. The diff is computed for reporting (so "Would add table indexes" output is preserved) but no index documents are written, so the IndexWorker never wakes up. This is safe because `wait_for_schema` is already gated on schema validation only for dry runs — index backfill was never meaningful to wait for in a dry-run context.

**4. `wait_for_schema` returns at `Validated`, not full backfill**

Since dry runs don't create index documents, `load_component_schema_status` short-circuits the index count when `dry_run: true`, returning `complete` as soon as the schema reaches `Validated`. This avoids a spurious wait on backfill that will never happen.

## What dry-run now does (and doesn't do)

| | Before | After |
|---|---|---|
| Schema committed to `_schemas` | ✓ always | ✓ always (needed for SchemaWorker validation) |
| Schema enforced on live writes | ✓ until next deploy | ✗ rolled back at end of dry run |
| Index documents created | ✓ IndexWorker backfills | ✗ never created |
| Index diff reported to CLI | ✓ | ✓ (via `get_index_diff`) |
| Function activation | ✗ | ✗ |

Schema validation against real data is preserved — that's the valuable part of Jordan's October 2025 intent, and this PR keeps it.

## Backwards compatibility

Old CLI versions that don't send `dry_run` in `StartPushRequest` or `WaitForSchemaRequest` get `false` via `#[serde(default)]` and see identical behaviour to today. New CLI with old backend: the flag is present in request bodies but ignored — same broken behaviour as today until backend is updated. The migration path is: ship backend, CLI follows.

### DIFF
--- a/crates/application/src/deploy_config.rs
+++ b/crates/application/src/deploy_config.rs
@@ -15,6 +15,7 @@ use common::{
     auth::AuthInfo,
     bootstrap_model::{
         components::definition::ComponentDefinitionMetadata,
+        push_mode::PushMode,
         schema::{
             SchemaMetadata,
             SchemaState,
@@ -46,6 +47,7 @@ use database::{
     BootstrapComponentsModel,
     IndexModel,
     OccRetryStats,
+    SchemaModel,
     Token,
     WriteSource,
     SCHEMAS_TABLE,
@@ -170,7 +172,11 @@ struct EvaluatedPushContents {
 
 impl<RT: Runtime> Application<RT> {
     #[fastrace::trace]
-    pub async fn start_push(&self, config: &ProjectConfig) -> anyhow::Result<StartPushResult> {
+    pub async fn start_push(
+        &self,
+        config: &ProjectConfig,
+        push_mode: PushMode,
+    ) -> anyhow::Result<StartPushResult> {
         let EvaluatedPushContents {
             app,
             auth_info,
@@ -182,7 +188,7 @@ impl<RT: Runtime> Application<RT> {
         } = self.evaluate_push_contents(config).await?;
 
         let schema_change = self
-            .handle_schema_change_in_start_push(&app, &evaluated_components)
+            .handle_schema_change_in_start_push(&app, &evaluated_components, push_mode)
             .await?;
         self.database
             .load_indexes_into_memory(btreeset! { SCHEMAS_TABLE.clone() })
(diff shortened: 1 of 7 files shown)


erquhart:
@zack-stoke thanks for this - we have an alternative fix in place and will release soon, let us know if you find any issues with it: https://github.com/get-convex/convex-backend/commit/ae88a47e3a0b74db3be16b4249b178412f53e4eb

PR #471, opened 2026-05-13
fix: Do not mark UseQueryResult as an internal export

leonekmi:
In https://github.com/get-convex/convex-backend/commit/0f8d1cb52b31dfa7c50b8986030104b9bf660437, the experimental object-form `useQuery_experimental` hook was exposed.
However, it appears that a previous version of that was reverted in https://github.com/get-convex/convex-backend/commit/b68d5d0dcbe57dfd169f8d43fa1f574b77bb202f, marking exports as internal.

This PR reverts the change in https://github.com/get-convex/convex-backend/commit/b68d5d0dcbe57dfd169f8d43fa1f574b77bb202f to export the return type, to let users infer the return type (e.g. when creating custom hooks) of `useQuery_experimental` without facing TS2742 (The inferred type of 'X' cannot be named without a reference to '{...}/convex/dist/esm-types/react/client'. This is likely not portable. A type annotation is necessary.) when using TypeScript.

----

### DIFF
--- a/npm-packages/convex/src/react/index.ts
+++ b/npm-packages/convex/src/react/index.ts
@@ -82,6 +82,7 @@ export {
   type MutationOptions,
   type ConvexReactClientOptions,
   type OptionalRestArgsOrSkip,
+  type UseQueryResult,
   ConvexReactClient,
   useConvex,
   ConvexProvider,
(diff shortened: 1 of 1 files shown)


dowski:
Thanks for flagging this and for the contribution. I attempted to include you as co-author on the commit but failed. In any event, the fix should go out in our next NPM release.

PR #472, opened 2026-05-15
Add eslint-plugin rule no-unexported-convex-function

akshatsinha0:
Added an ESLint rule to catch Convex function registrations that are declared but not exported from entry-point modules.

This prevents runtime dead code in `convex/` by flagging top-level `query`, `mutation`, `action`, `internalQuery`, `internalMutation`, and `internalAction` declarations that are never exported.

### DIFF
--- a/npm-packages/@convex-dev/eslint-plugin/src/lib/no-unexported-convex-function.ts
+++ b/npm-packages/@convex-dev/eslint-plugin/src/lib/no-unexported-convex-function.ts
@@ -2,7 +2,11 @@ import type { TSESTree } from "@typescript-eslint/types";
 import { AST_NODE_TYPES } from "@typescript-eslint/types";
 import { CONVEX_REGISTRARS, createRule, isEntryPoint } from "../util.js";
 
-type Options = [];
+type Options = [
+  {
+    additionalRegistrars: string[];
+  },
+];
 type MessageIds = "no-unexported-convex-function";
 
 /**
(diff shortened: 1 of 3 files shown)


npm-packages/@convex-dev/eslint-plugin/src/lib/no-unexported-convex-function.ts
```diff
@@ -0,0 +1,98 @@
...
+          parent.type === AST_NODE_TYPES.Program ||
+          (isExported && parent.parent?.type === AST_NODE_TYPES.Program);
+        if (!atTopLevel) return;
+
+        for (const declarator of node.declarations) {
+          if (declarator.id.type !== AST_NODE_TYPES.Identifier) continue;
+          const init = declarator.init;
+          if (!init || init.type !== AST_NODE_TYPES.CallExpression) continue;
+          if (init.callee.type !== AST_NODE_TYPES.Identifier) continue;
+          if (!CONVEX_REGISTRARS.includes(init.callee.name)) continue;
```
ianmacartney: does this work for custom functions - e.g. when they've made an `authMutation` instead of just `internalMutation` etc?

akshatsinha0: @ianmacartney, No not right now, but we can add a rule option for additional registrar names, eg, `{ additionalRegistrars: ["authMutation"] }`, because custom funcs can be named arbitrarily and inferring them would require import/export analysis.

_Needed to add tests as well!_

akshatsinha0: @ianmacartney Added support for custom wrapper names through an `additionalRegistrars` rule option in commit [`083809acc`](https://github.com/get-convex/convex-backend/pull/472/commits/083809acce58f1a6e3def6d9d841f405345f5936), so projects can configure names like `authMutation` without the rule guessing from imports or wrapper definitions. I also added RuleTester coverage for built-in registrars, configured custom registrars, unconfigured custom names, re-exported functions, and generated files.

PR #475, opened 2026-05-24
[RFC] Narrow query result types from literal equality predicates

ricglz:
## Summary

This is a prototype/RFC for carrying literal equality predicates from Convex index and search-index builders into the returned document type.

Today, a query like this still returns the full schema union for `visibility`:

```ts
const docs = await ctx.db
  .query("flashcardSets")
  .withIndex("by_visibility_and_createdAt", (q) =>
    q.eq("visibility", "public" as const),
  )
  .take(10);
```

Even though the index range requires `visibility === "public"`, `docs[number]["visibility"]` remains something like `"private" | "unlisted" | "public"`.

The concrete app motivation: in our app we currently need post-query guards like `isPublicFlashcardSet` immediately after `q.eq("visibility", "public")`, because Convex returns the full `Doc<...>` type instead of reflecting the invariant established by the query.

This PR makes that invariant visible to TypeScript for top-level literal equality predicates.

## What changed

- Add a carried result-document generic to `Query` and `OrderedQuery`, defaulting to the existing document type.
- Give `IndexRange` and `SearchFilter` a type-only carried result document.
- Make `IndexRangeBuilder.eq(...)` and `SearchFilterFinalizer.eq(...)` accumulate conservative top-level literal equality narrowing.
- Thread the narrowed document type through `withIndex(...)` and `withSearchIndex(...)` so query consumers return the narrowed document.
- Preserve runtime behavior. The implementation changes are generic annotations/casts around the existing builder objects and serialization paths.

## Deliberate scope limits

This is intentionally conservative for an RFC/prototype:

- Only top-level field paths narrow.
- Nested field paths like `metadata.visibility` do not narrow yet.
- Broad primitive variables, object equality values, and array equality values do not narrow.
- Non-literal/no-equality queries keep the existing full document type.

Those limits keep the type design readable and avoid trying to solve every field-path case in the first patch.

## Type tests

Added focused type tests covering:

- `withIndex(... q.eq("visibility", "public" as const))`
- chained index `.eq(...)` calls
- search index `.search(...).eq(...)`
- chained search filter `.eq(...)` calls
- preservation through `order`, `filter`, `first`, `unique`, `paginate`, and async iteration
- equality followed by range bounds
- no-narrowing behavior for non-literal variables, no equality, identity index ranges, nested paths, objects, and arrays
- optional fields, union document types, and branded Convex IDs

## Verification

Ran from `npm-packages/convex`:

```sh
npx --yes --package typescript@5.0.3 tsc -p tsconfig.type-tests.json --pretty false
```

Also checked:

```sh
git diff --check
```

The full package `tsc` command is blocked in this checkout because npm dependencies are not installed, but filtering its output showed no remaining `src/server` or `type-tests` errors.

### DIFF
--- a/npm-packages/convex/package.json
+++ b/npm-packages/convex/package.json
@@ -197,6 +197,7 @@
     "test": "vitest --silent",
     "test-not-silent": "vitest",
     "typecheck": "tsc --noEmit --emitDeclarationOnly false",
+    "typecheck:type-tests": "tsc -p tsconfig.type-tests.json",
     "test-esm": "node ./scripts/test-esm.mjs && ./scripts/checkdeps.mjs && ./scripts/checkimports.mjs",
     "compare-codegen": "node scripts/compare-codegen.mjs",
     "generateApiSpec": "npm run generateManagementApiSpec && npm run generateFunctionLogsApiSpec",
--- a/npm-packages/convex/src/server/data_model.ts
+++ b/npm-packages/convex/src/server/data_model.ts
@@ -1,4 +1,5 @@
 import { Value } from "../values/index.js";
+import { BetterOmit, Expand } from "../type_utils.js";
 import type { SystemIndexes } from "./system_fields.js";
 
 // Document Types  /////////////////////////////////////////////////////////////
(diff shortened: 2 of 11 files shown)


ianmacartney:
Really nice! This makes a lot of sense and seems like a natural fit. We might not have bandwidth short-term to shepherd this through, but will keep it in mind and hope to get to it soon. For now, if you get tests passing that'd be a good start and would enable making a prototype build to play with

ricglz:
Note on the current `Build Convex Backend` CI failure: the job is failing before it reaches the backend build or tests. `sccache --start-server` is being configured with the R2 cache settings, but this fork PR run does not receive the repository R2 secrets, so the access key is empty and sccache exits with:

```text
Credential access key has length 0, should be 32
```

I am leaving the workflow/action unchanged in this PR to keep it scoped to the query result type changes. The local TypeScript checks for this PR pass after the formatting fixes.

PR #477, opened 2026-05-27
feat: dashboard rename column / table tooling (#342)

### LINKED ISSUE #342: Feature Request: Renaming columns or tables simply
Apologies if this is duplicate but I can't find where someone mentioned this being an issue. 

Renaming columns or tables requires a flow like `add‑new‑field + migrate + remove‑old‑field` which can get rough when using an entire new terminology over the system. 

In my case, I named many things in my schema after the term `lead` but I need to change it to `client`.

mvanhorn:
## Summary

The data browser now offers a `Rename column...` action in the column context menu and a `Rename table...` action in the table sidebar overflow menu. Both open a shared dialog that generates a migration mutation plus a schema diff scaffold, copies it to the clipboard, and lets the user apply it through the normal `npx convex dev` flow.

## Why this matters

Issue #342 (5 reactions, 3 comments) describes the current workflow for a rename: edit `convex/schema.ts` to add the new field beside the old, hand-write a migration mutation, edit schema again to remove the old field, push twice. The dashboard is the natural place to scaffold that work.

This implementation does **not** apply the migration directly. The CLI-driven schema-push remains the authority. The dashboard generates text the user copies into their `convex/` folder. That keeps the safety model intact and stays consistent with how #439 (schema field order in the data browser, which landed earlier this year) limited itself to read-side dashboard polish.

The generated mutation matches the `@convex-dev/migrations` component pattern so it drops in cleanly for users who already use that component.

## Demo

Simulated demo of the column-rename flow, end to end. The actual dashboard render against a live backend wasn't captured (would have needed a full self-hosted setup); this mockup uses the same fonts, color tokens, and component shapes so reviewers can sanity-check the UX shape:

demo

MP4 version: https://files.catbox.moe/bbg1in.mp4

The scaffold text the dialog generates was exercised against four representative inputs (column with backfill, table with backfill, backfill-off, non-identifier column name). Sample output for `posts.title -> posts.heading`:

```typescript
// convex/migrations.ts
import { internalMutation } from "./_generated/server";

export const renamePostsTitleToHeading = internalMutation({
  args: {},
  handler: async (ctx) => {
    const docs = await ctx.db.query("posts").collect();
    for (const doc of docs) {
      const value = doc.title;
      if (value !== undefined) {
        await ctx.db.patch(doc._id, { heading: value, title: undefined });
      }
    }
  },
});

// convex/schema.ts
posts: defineTable({
-  title: /* existing validator */,
+  heading: /* existing validator */,
  // ...
}),
```

The schema diff uses `/* existing validator */` as a placeholder so the user preserves whatever validator type the field had, rather than the dashboard hardcoding `v.string()`. The dialog text spells this out.

For table renames, the generated mutation does `insert + delete`, which assigns new document IDs. The dialog renders a `<Callout variant="error">` above the preview warning that `v.id()` references in other tables need to be updated in the same migration, and the generated mutation repeats the warning in a comment block at the top of the code:

```typescript
// WARNING: This migration assigns new document IDs to renamed rows.
// References to this table (v.id("user_profiles")) from other tables will break.
// Before running: either rename columns instead of the whole table,
// or update all referencing fields in the same migration.
```

## Changes

- New `renameScaffold.ts` (pure-TS, no React) exporting `generateColumnRenameScaffold` and `generateTableRenameScaffold`. Identifier-escape helpers handle non-identifier names; PascalCase the mutation export name.
- New `RenameDialog.tsx` wraps the design-system `Modal`/`TextInput`/`Checkbox`/`Callout` primitives. Reuses `validateConvexIdentifier` from the existing helpers. Per-section `CopyButton`s plus a combined "Copy mutation + schema diff" primary button. The table-mode `<Callout variant="error">` is the load-bearing UX guard against the id-reference footgun.
- `TableContextMenu.tsx` gets a new `onClickRenameColumn` prop and renders a `Rename column...` item (gated off the `*select` sentinel column).
- `DataOverflowMenu.tsx` gets a new `onClickRenameTable` prop and renders a `Rename table...` item between `Indexes` and `Metrics`.
- `Table.tsx` and `DataToolbar.tsx` thread the prop through to their `TableContextMenu` and `DataOverflowMenu` call sites.
- `DataContent.tsx` owns the `renameTarget` state, the rename callbacks, and the `<RenameDialog>` render.

## Testing

`npx prettier --check` passes on all touched files. The scaffold generator was exercised against the four input shapes (column rename with and without backfill, table rename with backfill, and a non-identifier column name); each produced expected output.

I did not add a unit-test file in this change. `npm-packages/dashboard-common` has Jest in its dev dependencies but no `jest.config` or test entry, so a new test would have been the first one in the package; rather than add the test-runner wiring in the same PR, I left that for a follow-up so the rename feature itself stays narrowly scoped.

I did not run the dashboard against a live backend. The demo above is a HyperFrames mockup using the same color tokens and component shapes; happy to capture a real screenshot reel from a self-hosted setup if that would help review.

## Scope notes for reviewers

Following the CONTRIBUTING.md guidance to flag the approach on larger changes so the team can early-redirect if the UX shape isn't what you want:

- The dialog generates text and copies it. No new admin RPC, no change to the schema-push protocol, no CLI changes.
- Column rename is the common case; table rename ships behind the warning callout because of the id-reference issue.
- The `@convex-dev/migrations` pattern is the target so component users get a drop-in. If you'd prefer a different mutation shape, the change is localized to `renameScaffold.ts`.

Fixes #342


ianmacartney:
Pretty cool! What do you think about making this a TUI under a CLI command for the migrations component? 
```sh
npx @convex-dev/migrations create [--rename-field] ...
```
That seems like the more idiomatic place to put it, and it could then be used by agents similarly.
The CLI could also help with scaffolding the component setup if you hadn't installed / added it to your app yet.

mvanhorn:
Agreed that the component CLI is the more idiomatic home - I built it: get-convex/migrations#59. npx @convex-dev/migrations create supports --rename-field table.old=new and --rename-table old=new non-interactively (agent-friendly, with --json summaries), falls back to prompts with no flags, and detects a missing component setup and prints the config steps instead of failing. Happy to close this dashboard PR in favor of that one, or keep both if you see value in the dashboard affordance.

ianmacartney:
Thanks for the PR! I haven't had bandwidth to review it in depth yet, but you can try it out with `npm i https://pkg.pr.new/@convex-dev/migrations@59`. I think that is the more promising direction vs. the dashboard tooling.

mvanhorn:
Thanks. Agreed the migrations CLI direction feels cleaner; happy to focus on #59 and fold anything useful from this into it. If that lands, this one can close in its favor.

PR #480, opened 2026-05-29
fix(common): avoid panic on integer overflow / divide-by-zero in query filter arithmetic

sravan27:
## Summary

Expression arithmetic on `Int64` values in the query filter evaluator (`crates/common/src/query.rs`, `binary_arithmetic` + `Neg`) uses raw `+ - * / %` and unary `-`. Under the release profile's `overflow-checks` these **panic** on:

- division / remainder by zero (`$div`, `$mod`)
- integer overflow: `i64::MIN / -1`, `i64::MIN % -1`, `add`/`sub`/`mul` overflow, and `-i64::MIN`

Filter expressions are built from user-controlled query input (`q.add` / `q.div` / … in the client query builder) and there is no `catch_unwind`, so a query such as `q.eq(q.div(q.field("a"), 0), 1)` can panic the worker rather than returning an error.

It's also an internal inconsistency: `Float64` operands already fail soft (`inf`/`NaN`), and type mismatches already return a graceful `EvalError` — only integer arithmetic aborts.

## Fix

Use `checked_add` / `checked_sub` / `checked_mul` / `checked_div` / `checked_rem` (and `checked_neg`), returning the same `ErrorMetadata::bad_request("EvalError", …)` on overflow / divide-by-zero that the type-mismatch path already uses. In-range arithmetic is unchanged.

## Notes

The panics were reproduced with a standalone harness exercising these operators under `overflow-checks`. `rustfmt` (edition 2024, repo config) applied cleanly. Opened for review so CI can run the full Rust suite; happy to add a unit test in `query.rs` if you point me at the preferred location.

### DIFF
--- a/crates/common/src/query.rs
+++ b/crates/common/src/query.rs
@@ -719,15 +719,22 @@ fn binary_arithmetic<I, F>(
     do_floats: F,
 ) -> anyhow::Result<ConvexValue>
 where
-    I: FnOnce(i64, i64) -> i64,
+    I: FnOnce(i64, i64) -> Option<i64>,
     F: FnOnce(f64, f64) -> f64,
 {
     let l = l_expr.eval(environ)?;
     let r = r_expr.eval(environ)?;
 
     let result = match (&l.0, &r.0) {
-        (Some(ConvexValue::Int64(l)), Some(ConvexValue::Int64(r))) => {
-            val!(do_ints(*l, *r))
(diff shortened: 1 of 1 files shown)


Nicolapps:
Thank you! I made a fix based on your report that I just merged (cbae5af0a42b85284f0a810cb023acb9dbb67a37)

Issue #481, opened 2026-05-31
`npx convex env list` prints secret values by default


drake-nathan:
### Summary

`npx convex env list` prints environment variable values, including plaintext secrets,
by default. There's no way to list just the names, and no redaction. That makes it really
easy to leak secrets into places they shouldn't end up, most commonly agent and CLI
transcripts.

### Repro

```bash
⏺ Bash(npx convex env list 2>&1 | tr -d '\000')
  ⎿  ANTHROPIC_API_KEY=***********
     GOOGLE_PERSONAL_CLIENT_ID=*******************
     GOOGLE_PERSONAL_CLIENT_SECRET=******************
     … +4 lines (ctrl+o to expand)
  ⎿  Allowed by auto mode classifier
```

(values redacted here, but they print in full)

CLI version: 1.39.1

### Why I think this is worth changing

Listing what's configured feels like a harmless thing to do, but right now it dumps every
secret as a side effect. Those two things should be separate.

The bigger issue for me is agents. Coding agents run `convex env list` constantly to check
config, and every time they do, the values land straight in the transcript and model
context. This has happened to me half a dozen times now. The first time, my agent actually
stopped and told me to rotate my tokens, so even the agents don't expect this command to
leak secrets.

I've read the [Agent Mode docs](https://docs.convex.dev/cli/agent-mode) and the suggestion
to give agents isolated deployments. That's a good fit for cloud agents, but it doesn't
really cover the common case of a person working with a local agent on their own dev
deployment. It's also opt-in, and the default still prints everything. With all the
security issues happening in the ecosystem lately, I think hardening the default here is
worth doing.

### Proposed changes (non-breaking)

1. Default to names only, with values behind an explicit flag like
   `convex env list --show-values` or `--reveal`. If flipping the default outright is too
   aggressive, you could gate it behind a config flag so it rolls out without breaking
   anyone's scripts.
2. At a minimum, redact by default (`SOME_API_KEY=sk_live_****`) with a flag to reveal.
3. Optionally, auto-redact when there's no TTY, so it's safe when an agent or script runs
   it non-interactively.

Either way, `convex env get NAME` stays as the intentional "I actually want this value" path.

I'm more than happy to put up a PR for whatever direction the team prefers.

ianmacartney:
Unfortunately this would be a breaking change. One common pattern is to
pipe `npx convex env list` to a file or another command. There's a few
options, but nothing seems like a slam dunk, so it's not just a matter of
[...]

PR #482, opened 2026-06-08
fix(cli): add --names-only flag to env list to avoid printing secret values

mvanhorn:
## Summary
`npx convex env list` prints every variable as `NAME=value`, including plaintext secrets. This adds an opt-in `--names-only` flag that lists just the variable names, so you can check what's configured without the values landing in terminal scrollback, CI logs, or an agent's transcript. Default output is unchanged.

## Why this matters
The reporter (#481) hit this through coding agents: they run `convex env list` constantly to check config, and every run dumps secrets into the transcript and model context. One of their agents stopped and told them to rotate their tokens. `convex env get NAME` already exists for the intentional "I want this value" case, but there was no way to answer "which variables are set?" without printing all of them.

This is deliberately the smallest, non-breaking step:
- `--names-only` is opt-in. `convex env list` with no flag is byte-for-byte unchanged, so existing `convex env list > .env.convex` and pipe-to-command workflows behave exactly as before (the concern you raised on the issue).
- In names-only mode the value is never formatted, so the `Warning (NAME): ...` line (which is derived from the value) is suppressed too. No value-derived content reaches any output channel.

It does not change the default. #481 floats hardening the default (names-only by default, redact-by-default, or auto-redact when non-TTY). Since you flagged that none of those is a slam dunk and the non-TTY case is the pipe-to-file pattern you want to preserve, I left the default alone here. Happy to follow up with whichever default direction you prefer. And `--names-only` is just one option name; if you'd rather call it `--reveal`/`--show-values` (inverting the default), say the word and I'll rename before you spend review time on it.

## Testing
Added a unit test (`src/cli/lib/env.test.ts`) that runs `envList` with a fake backend and `{ namesOnly: true }`, including a value with `'` and `#` (which normally triggers a dotfile warning), and asserts the output is names-only with no values and no `Warning (` lines.

Refs #481

---

### DIFF
--- a/npm-packages/convex/src/cli/env.ts
+++ b/npm-packages/convex/src/cli/env.ts
@@ -164,11 +164,11 @@ const envRemoveCmd = new Command("remove")
   });
 
 const envListCmd = new Command("list")
-  .summary("List all variables")
+  .summary("List all variables and their values")
   .description(
     [
-      "• List all variables: `npx convex env list`",
-      "• List only variable names: `npx convex env list --names-only`",
+      "• List all variables and their values: `npx convex env list`",
+      "• List only variable names (no values): `npx convex env list --names-only`",
       "• Save all variables to a file: `npx convex env list > .env.convex`",
       "• Append to a file: `npx convex env list >> .env.convex`",
     ].join("\n"),
@@ -207,7 +207,8 @@ export const env = new Command("env")
       "• Set interactively: `npx convex env set NAME`",
       "• Set multiple from file: `npx convex env set --from-file .env`",
       "• Unset a variable: `npx convex env remove NAME`",
-      "• List all variables: `npx convex env list`",
+      "• List all variables and their values: `npx convex env list`",
+      "• List only variable names (no values): `npx convex env list --names-only`",
       "• Print a variable's value: `npx convex env get NAME`",
       "",
       "By default, this sets and views variables on your dev deployment.",
--- a/npm-packages/convex/src/cli/env.ts
+++ b/npm-packages/convex/src/cli/env.ts
@@ -168,14 +168,21 @@ const envListCmd = new Command("list")
   .description(
     [
       "• List all variables: `npx convex env list`",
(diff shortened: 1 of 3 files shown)


ianmacartney:
Sounds reasonable!
We should update the `npx convex env --help` text to make it clear that `env list` will list the values. Right now it just says "List all variables"
Currently they might guess to run `env list` before doing `--help` to discover `--names-only` - do you recall if your agent looked up the env list help before running it?
I suspect `npx convex env list-names` might be enough for them to guess that `env list` returns more than just names. But flag also works, and if the docs / help text suggests it, I think we'll be good. 

Also a future where some env vars are marked as explicit secrets could be a good time to require passing --include-secrets in `list` to show those

mvanhorn:
Done, pushed in 80d21a3.

- `convex env list` summary now reads "List all variables and their values" (was "List all variables"), and the same wording is in the `env --help` bullet.
- I also added the `--names-only` variant directly to the parent `env --help` listing, so an agent reading `env --help` sees it before ever running `env list`. That covers the "they guess `env list` before checking `--help`" case you raised: `--help` itself now advertises both.

On whether the reporter's agent checked `env list --help` first: I'm the PR author here, not the original reporter from #481, so I can't speak to that agent's exact lookup order. But I agree the help text is the right lever, if `env --help` spells out the values-vs-names distinction, an agent has what it needs without a trial run.

The `--include-secrets` idea for explicitly-marked secrets sounds like a clean next step once that metadata exists; happy to do it separately so this PR stays minimal.

(Also added the CLA confirmation line to the description, so that check should clear on the next run.)

Nicolapps:
Thank you for your PR! Could you please also make sure the `envDefaultList` command in `envDefault.ts` is also updated the same way?

ianmacartney:
I'll update this in our internal repo, thanks @mvanhorn

ianmacartney:
landed in 53e832c1095ae28effceab95d98f17dfbd1a3f29 thanks!
It will go out in the next CLI version

Issue #476, opened 2026-05-26
Delete All Records in DB/Reset database


JJosephttg:
I don't know if there is any documentation around this or not, but I am having a hard time trying to figure out how to reset the database without completely deleting the docker volume. Ideally there should be an API or something that allows me to programmatically/through the CLI to be able to wipe out all data. 

I understand I can go to each table and clear out the data, but that doesn't make sense from a maintenance perspective. I find it would be insanely tedious to add to a script just to remove data from each individual table that ever exists. Especially when it comes to plugins whose tables you do not maintain. 

Is there a quick and easy way to clear out all my data so that I have empty tables, or am I out of luck?

Personally I seed my data using internal mutation scripts

satishvudata98:
For Convex specifically, there is no built-in “reset database” API today that wipes all tables while preserving schema the way traditional SQL systems sometimes support.

For production deployments, Convex May be intentionally does not expose a global “DROP ALL DATA” style API because:

- tables are schema-less/dynamic

The common approaches for resetting database are basically:

- Delete the deployment / recreate it
- Clear tables manually via mutations
- Use the dashboard to delete documents per table
- Reset the local dev instance

ianmacartney:
You can also use snapshot import to clear tables by replacing the table with an empty one. My one-liner for this from the CLI is:

one table:
```
npx convex import --table $tableName --replace --format jsonLines /dev/null -y
```
All tables in the app:
```
for tableName in `npx convex data`; do npx convex import --table $tableName --replace -y --format jsonLines /dev/null; done
```
All tables in a component:
```
for tableName in `npx convex data --component $component`; do npx convex import --component $component --table $tableName --replace -y --format jsonLines /dev/null; done
```

mvanhorn:
Documented the clear/reset workflow in #483 (new self-hosted/advanced/clearing_data.md, linked from the self-hosted README). It captures @ianmacartney's one-liners from this thread for clearing one table, all tables, and a component, with an export-first backup step matching the other advanced docs. No new command, just documenting the supported path.

PR #483, opened 2026-06-09
docs(self-hosted): document clearing and resetting data

mvanhorn:
## Summary
Adds a self-hosted docs page explaining how to clear or reset data without deleting the Docker volume, by replacing tables with an empty snapshot import. Linked from the Advanced Configuration list.

## Why this matters
The reporter (#476) couldn't find any documentation for resetting a self-hosted database and assumed there was no supported way short of deleting the Docker volume. The capability already exists via `npx convex import --replace`, and @ianmacartney posted the exact one-liners in the thread (one table, all tables, all tables in a component), but a GitHub comment isn't discoverable later. This moves that answer into the docs so the next person searching "reset self-hosted convex database" finds it. No new command, no behavior change.

It follows the existing self-hosted docs convention: the page leads the destructive commands with an `npx convex export` backup step, matching upgrading.md and s3_storage.md, and warns to confirm `CONVEX_SELF_HOSTED_URL` before an all-tables clear.

## Testing
Docs-only, prettier-checked. The commands are the ones from the issue thread (`npx convex data` lists bare table names to stdout, and `npx convex import --replace --format jsonLines /dev/null` clears a table).

Refs #476

### DIFF
--- a/self-hosted/README.md
+++ b/self-hosted/README.md
@@ -119,6 +119,7 @@ e.g., instead of setting `CONVEX_DEPLOY_KEY`, you'll need to set
 - [Hosting on your own servers](./advanced/hosting_on_own_infra.md)
 - [Running the database on Postgres or MySQL](./advanced/postgres_or_mysql.md)
 - [Storing files in S3 instead of local filesystem](./advanced/s3_storage.md)
+- [Clearing or resetting data](./advanced/clearing_data.md)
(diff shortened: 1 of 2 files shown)


ianmacartney:
Thanks! Just submitted it internally

ianmacartney:
landed in 9deb1352f4a313ed8db4f5c4e37acfb320c351ff

Issue #418, opened 2026-03-26
Blank page after deleting the account


typed-sigterm:
Deleting account redirects to `https://api.workos.com/user_management/sessions/logout?session_id=session_***&return_to=https%3A%2F%2Fdashboard.convex.dev%2Flogin`, a blank page.

ianmacartney:
Can you give some more details here? Were you deleting your own Convex account @typed-sigterm ?

typed-sigterm:
I registered another account by mistake. After I deleted the account, it redirects me to `https://api.workos.com/user_management/sessions/logout?session_id=session_***&return_to=https%3A%2F%2Fdashboard.convex.dev%2Flogin` and nothing shows. Maybe it should be a Goodbye screen or redirects to homepage I think, or users may be confused about what happened.

manaskumar3003:
can i work on fixing this issue?

ianmacartney:
Yes, PRs welcome for this, assuming it's a simple fix. If it is a
configuration issue on our end with our WorkOS setup, a PR may not suffice.

atrakh:
Fixed in https://github.com/get-convex/convex-backend/commit/c0c4c077618a0ec07fb5d77e5880d14507ea644f

PR #485, opened 2026-06-11
fix(dashboard): bypass WorkOS redirect after account deletion to prev…

Adityakk9031:
Description
This PR addresses issue #418, where deleting a user account leads to a blank page on the WorkOS domain.

Root Cause
When a user requests account deletion, the backend deletes the user profile and invalidates all active sessions within WorkOS.
After the deletion call completes, the dashboard attempts to clear local state and redirects the browser to the Next.js API logout handler (/api/auth/logout) to erase the HttpOnly session cookie (wos-session).
The API logout handler loads the session and attempts to fetch a WorkOS-managed logout URL using session.getLogoutUrl({ returnTo: ... }).
Because the session has already been deleted in WorkOS (via the user deletion flow), redirecting to the WorkOS logout URL with that session ID causes WorkOS to fail/hang on a blank page instead of redirecting back to the login page.
Solution
Added a local=true query parameter to the /api/auth/logout Next.js API route. When specified, it deletes the wos-session cookie and immediately redirects back to /login without invoking the external WorkOS logout URL.
Updated the confirmation action in profile.tsx to redirect the user to /api/auth/logout?local=true once the deleteAccount API request succeeds.
Changes
npm-packages/dashboard/src/pages/api/auth/logout.ts
Added check for req.query.local === "true" in the session-validation conditional to redirect to /login early.
npm-packages/dashboard/src/pages/profile.tsx
Updated redirection in the confirmation callback from /api/auth/logout?returnTo=/api/auth/login to /api/auth/logout?local=true.

### DIFF
--- a/npm-packages/dashboard/src/pages/api/auth/logout.ts
+++ b/npm-packages/dashboard/src/pages/api/auth/logout.ts
@@ -17,7 +17,7 @@ export default async function handler(
 
     res.setHeader("Set-Cookie", deleteSessionCookie());
 
-    if (!session) {
+    if (!session || req.query.local === "true") {
       return res.redirect("/login");
     }
 
--- a/npm-packages/dashboard/src/pages/profile.tsx
+++ b/npm-packages/dashboard/src/pages/profile.tsx
@@ -92,8 +92,7 @@ export function Profile() {
                         document.cookie = "";
                         window.localStorage.clear();
                         await deleteAccount();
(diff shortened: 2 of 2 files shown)


atrakh:
Fixed in https://github.com/get-convex/convex-backend/commit/c0c4c077618a0ec07fb5d77e5880d14507ea644f, thank you for the contribution

PR #491, opened 2026-06-22
Cache subqueries within a query execution (ctx.runQuery memoization)

adelin-b:
## What

When a query calls another query via `ctx.runQuery`, memoize the subquery's result for the rest of that top-level query execution, keyed by `(component, udf_path, args)`. Repeated identical subquery calls are served from an in-transaction cache instead of re-executing.

Today nested queries run as subtransactions inside the parent transaction (`crates/isolate/.../udf/async_syscall.rs::run_udf`) with **no** memoization — they fully re-execute on every call. The top-level query result cache (`crates/application/src/cache`) only caches top-level queries.

## Why it's correct

A query tree is **read-only**: a query can only call queries, and queries can't write, so the whole tree executes against a single **immutable transaction snapshot**. The same `(component, path, args)` therefore always yields the same result.

Reactivity is preserved **for free**: the first execution's reads are already merged into the parent transaction's `TransactionReadSet`, and subtransactions never roll back reads (`SubtransactionToken` savepoints writes/index/tables/registries, but not `reads`). Skipping a later identical call loses no read dependency, so the parent query is still invalidated whenever any data the subquery read changes.

## Guards (see `crates/database/src/subquery_cache.rs` module docs)

- **Query parent only** — mutations can write between two subquery calls; **system UDFs excluded** (they thread a pagination journal through nested calls).
- **Only `Ok` results** — a subquery that errors re-executes next call.
- **No observed randomness** — each `ctx.runQuery` advances the parent RNG and seeds the child independently; the single draw is mirrored on a cache hit via the shared `UdfPhase::draw_nested_rng_seed` helper, so the parent RNG stream is identical with or without memoization.
- **No audit logs** — a cache hit doesn't re-emit them.
- `observed_identity` / `observed_time` are **replayed on a hit** so the top-level query result cache key stays correct.
- Bounded by `MAX_SUBQUERY_CACHE_ENTRIES` (default 1024) per transaction; freed with the transaction.

## Behavior change

A subquery called N times with identical args now **logs and meters its reads once, not N times**. This is a strict reduction — the reads only physically happened once anyway.

## Scope

This is the in-transaction (single-execution) scope. A future cross-request subquery cache — reusing results across separate top-level queries and reactive recomputations — would live in the application layer alongside `CacheManager` and needs independent per-subquery read-set tokens. The new module is intentionally kept simple to keep that future work decoupled.

## Files

- `crates/database/src/subquery_cache.rs` (new) — `SubqueryCache` + key/value types + unit tests
- `crates/database/src/transaction.rs` — `subquery_cache` field + accessors
- `crates/database/src/lib.rs` — re-exports
- `crates/common/src/knobs.rs` — `MAX_SUBQUERY_CACHE_ENTRIES`
- `crates/isolate/src/environment/udf/async_syscall.rs` — memoization in `run_udf`
- `crates/isolate/src/environment/udf/phase.rs` — `draw_nested_rng_seed` / `advance_nested_rng`
- `crates/isolate/src/metrics.rs` — `log_subquery_cache` hit/miss counter

## Testing

`cargo test -p database subquery_cache` — 5 unit tests pass (insert/get, distinct keys, entry cap, observed-flag round-trip, key equality/hash). `database` and `isolate` compile; clippy clean on the changed crates.

> Note: the isolate JS-fixture integration harness (`npm-packages/tests/udf-tests/convex`) isn't present in the OSS checkout, so end-to-end fixture tests for the `run_udf` path aren't included here. The RNG one-draw invariant is enforced structurally (both paths share `draw_nested_rng_seed`) rather than by a test.


ianmacartney:
Did you take into account pending writes that the query can observe?
Eg calling a mutation on a component then a query - the tx will read its writes, even in nested udfs

adelin-b:
Good catch — fixed in 83c80ac.

Full disclosure: this PR and this review response were developed with heavy AI assistance, so I want to be transparent about *how* I checked the fix rather than just asserting it's correct. Please sanity-check the claims below against your own knowledge of the codebase — I may have misread something.

### The bug + fix

A nested query reads the **same** transaction as its caller, including writes already buffered in it (read-your-writes), and this memo cache lives on the `Transaction`, so it would otherwise outlive a write:

```
mutation M:
  runQuery(Q)        // Q -> runQuery(SubQ): reads D=v1, memoized on the tx
  db.patch(D, v2)    // pending write in the SHARED tx
  runQuery(Q)        // Q -> runQuery(SubQ): stale memo hit -> v1, must be v2
```

Fix: gate memoization on `Transaction::is_readonly()` (plus the existing query-context / non-system checks). A query tree never writes → full caching; a mutation stops memoizing the moment it buffers a write, so the second `runQuery(Q)` re-executes and sees `v2`. This also lines up with an existing convention — `Transaction::into_token_and_index_cache()` already gates the cross-transaction `TimestampedIndexCache` extraction on `is_readonly()`.

### Adversarial sweep for *other* spots like this

After the fix I ran an adversarial pass over the codebase looking for the same class of issue: any in-transaction cache / memoization / snapshot reuse that could observe pending writes, serve a stale value across a write, or skip determinism/observation replay. Here's what I found and my verdict for each — **corrections very welcome**:

| Site | Scope | Why it's safe (my reading) | Verdict |
|---|---|---|---|
| **SubqueryCache** (this PR) | in-tx | now gated on `is_readonly()` | **fixed** |
| Transaction index read path (`transaction_index.rs::range_no_deps`) | in-tx | merges `database_index_updates` (pending writes) live on every read; no per-read cache | safe |
| `PreloadedIndexRange` (`preloaded.rs`, env vars) | in-tx | read-only use; mutations don't reuse the preloaded range | safe |
| `TimestampedIndexCache` (`database_index_snapshot.rs`) | cross-tx | extracted only at read-only commit, after reads, never reused within the writing tx (`is_readonly()` gated) | safe |
| Top-level query result cache (`application/src/cache`, `CacheManager`) | cross-request | read-set `Token` refresh against the requested ts; identity/journal in key | safe |
| `InMemoryIndexCache` (`function_runner`) | cross-request | keyed by `last_modified` timestamp → write bumps the key | safe |
| Module caches (app + funrun) | cross-request | content-addressed by `sha256` (+ deployment) → immutable | safe |
| Snapshot metadata caches (`table_mapping` / `by_id_indexes` / `component_paths`) | cross-request | keyed by snapshot `Timestamp` | safe |
| `ListSnapshotTableIteratorCache` (`database.rs`) | per-snapshot | exact key match on (snapshot, tablets, index, cursor) | safe |

Net: the subquery cache was the **only** spot that reused a value incorporating a transaction's pending writes; everything else either reflects pending writes live on each read, is read-only, or is keyed by a timestamp/sha/read-set-token that a write moves off of.

I also checked the non-write determinism dimensions for the subquery cache (identity, time, randomness, component, FunctionHandle-vs-ref, args byte-equality, snapshot queries, pagination journal, errors, audit logs, actions) — all either keyed, replayed, or unreachable given the gates; happy to expand on any.

**Questions for the team:** does this match how you think about read-your-writes and cache invalidation in the transaction layer? And are there reuse sites I've missed that you'd want examined (e.g. search/vector index snapshots, text-search caches, scheduler/cron state, component-registry reuse)?

ianmacartney:
What about Mutation -> Query -> Query, will it think it's in a readonly context in the nested query, even though there are pending writes being applied?

adelin-b:
No — a nested query in a mutation tree is not treated as read-only.

`is_readonly()` is `writes.is_empty()`, and entering a nested UDF calls `begin_subtransaction()` → `NestedWrites::begin_nested()`, which seeds each nested layer with a **clone of the parent's pending writes** (`writes.rs:99`). So a mutation's buffered write stays visible (`is_empty() == false`) at any depth → memoization stays off.

Probed it as hypotheses (all came back "handled"); tests in 6f8ef0c (`cargo test -p database nested_writes_tests`):

| Case | Stale memo? |
|---|---|
| Mutation → Query → Query (your case) | no — write visible at every nesting level |
| committed nested write → later query | no — `commit_nested` keeps it |
| rolled-back nested write | no — `rollback_nested` restores read-only |
| pure Query → Query | n/a — never writes |

One caveat: `is_readonly()` watches *document* writes; it's complete only because every mutation state change (insert/patch/delete, table create, scheduler) also writes a system-table document. If you'd rather not lean on that, I can have the gate check the sibling write-sets directly — is there any mutation-side change that doesn't touch `writes`?

PR #500, opened 2026-07-05
fix(node-executor): prevent process.env race under concurrent Node actions

### LINKED ISSUE #492: Self-hosted: Node action env vars intermittently `undefined` under concurrency (runWithEnvironmentVariables mutates global process.env)
## Summary

On self-hosted Convex, Node actions (`"use node"`) intermittently fail to read configured deployment environment variables: `process.env.MY_VAR` is `undefined` even though the variable is set. It is **load-dependent** — it appears when several Node actions run **concurrently** (e.g. a client uploading a batch of files where each file triggers an action), while the same action succeeds the rest of the time.

We hit this in production as a flaky `S3_BUCKET_PHOTOS is not set` during batch photo uploads: most files upload fine, a few fail with the env var missing.

## Root cause

[`runWithEnvironmentVariables`](https://github.com/get-convex/convex-backend/blob/7055c85c08b70703d0696da6f8e77b5165b3d125/npm-packages/node-executor/src/executor.ts#L75-L118) wraps every Node action invocation like this:

```ts
const savedEnv = process.env;
process.env = sanitized;             // wholesale replace of the GLOBAL object
for (const e of envs) process.env[e.name] = e.value;
try {
  return await fn(envHash);
} finally {
  process.env = savedEnv;            // restore — drops the deployment vars
}
```

`process.env` is a single process-global object, but the executor runs **multiple invocations concurrently in one process** — only `globalConsoleState` / `globalDevConsole` / `globalSyscalls` are isolated per-invocation via `AsyncLocalStorage`; `process.env` is not. So with two overlapping invocations A and B:

1. **A**: `savedEnv = process.env` (original, no user vars); `process.env = { sanitized + user vars }`; `await fn(...)` (yields the event loop).
2. **B** starts while A is suspended and runs.
3. **A**’s `await` resolves → `finally` → `process.env = savedEnv` (**the original, without user vars**).
4. **B** is still running and now reads `process.env.MY_VAR` → `undefined`.

Because it depends on the exact interleaving of the `finally` blocks, it is intermittent and becomes more frequent as concurrency increases.

## Reproduction

A `"use node"` action that reads a configured env var after an `await`, invoked concurrently:

```ts
"use node";
import { action } from "./_generated/server";

export const readEnv = action({
  args: {},
  handler: async () => {
    await new Promise((r) => setTimeout(r, 50)); // any await / yield
    if (!process.env.MY_VAR) throw new Error("MY_VAR is undefined");
    return "ok";
  },
});
```

Set `MY_VAR`, then fire ~10–20 calls concurrently (`Promise.all`). A subset throw `MY_VAR is undefined`.

## Environment

- Self-hosted `convex-backend` (observed on the image dated 2026-04-30; the code path is unchanged on `main` at the time of writing).
- Single-node self-hosted deployment, Node runtime actions.

## Suggested fixes

1. **Minimal (ref-count):** only swap `process.env` on the first concurrent invocation and restore it on the last. Correct under the invariant that concurrent invocations in one executor process share the same env set (one process serves one deployment). I opened a draft PR with this approach.
2. **Fully isolated:** resolve `process.env` per invocation via `AsyncLocalStorage` (consistent with how syscalls/console are already isolated). More involved — needs a `process.env` Proxy and care around the Node.js 24 AWS-env workaround in the `finally` — but it does not rely on the same-env invariant.

Happy to iterate on whichever direction you prefer.

tsushanth:
Fixes #492.

## Root cause

`runWithEnvironmentVariables` does a wholesale swap of `process.env` per invocation, but the executor runs **multiple invocations concurrently in one process**. When invocation A's `finally` fires while invocation B is suspended at an `await` inside `fn()`, A restores `savedEnv` (the pre-sanitized Lambda env) and B subsequently reads `undefined` for every user-configured deployment variable. No exception is thrown — the failure is silent and load-dependent.

## Fix

Ref-count concurrent invocations. Perform the env swap only on the **first** entry (count `0→1`) and restore only on the **last** exit (count `1→0`).

This relies on the invariant that all concurrent invocations in one executor process serve the same deployment and therefore share an identical env set — consistent with the existing architecture where syscalls and console are already isolated per-invocation via `AsyncLocalStorage` but env variables are not.

The Node 24 AWS Lambda workaround (restore `process.env` so Lambda's own vars are present on a subsequent cold-start invocation) is preserved: the restore fires exactly once, on the last concurrent exit.

## Changes

- `npm-packages/node-executor/src/executor.ts`: add `activeEnvInvocations` counter + `savedLambdaEnv` snapshot; gate the swap and restore on counter transitions `0→1` / `1→0`.

### DIFF
--- a/npm-packages/node-executor/src/executor.ts
+++ b/npm-packages/node-executor/src/executor.ts
@@ -72,37 +72,49 @@ export function setupGlobals(modulePath: string) {
 
 let numInvocations = 0;
 
+// Tracks concurrent runWithEnvironmentVariables calls. All concurrent
+// invocations in a single executor process serve the same deployment, so they
+// share the same env set. We only swap process.env on the first concurrent
+// entry and restore it on the last exit, preventing the race where one
+// invocation’s finally-block restores the pre-sanitized env while a sibling
+// invocation is still running and reading its user-defined vars.
+let activeEnvInvocations = 0;
+let savedLambdaEnv: NodeJS.ProcessEnv | null = null;
+
 async function runWithEnvironmentVariables<T>(
   envs: EnvironmentVariable[],
   fn: (envHash: string) => Promise<T>,
 ): Promise<T> {
-  const savedEnv = process.env;
+  if (activeEnvInvocations === 0) {
+    savedLambdaEnv = process.env;
 
-  // AWS Lambda populates a number of environment variables, like Lambda version,
-  // handler name, session, AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, etc. We
-  // don't want to expose any of that. Only expose variables that are common
-  // between local Node.js and AWS Lambda.
-  //
-  // Note: This sanitization is for consistency between environments, not for security.
-  // While user code can still access underlying environment variables through
-  // other means, sensitive variables (such as AWS credentials) are protected
-  // through restrictive IAM policies that limit what the Lambda function can do
-  // with those credentials.
-  const allowedEnvs = ["PATH", "PWD", "LANG", "NODE_PATH", "TZ", "UTC"];
-  const sanitized: { [name: string]: string } = {};
-  for (const name of allowedEnvs) {
-    const value = process.env[name];
-    if (value !== undefined) {
-      sanitized[name] = value;
+    // AWS Lambda populates a number of environment variables, like Lambda
+    // version, handler name, session, AWS_ACCESS_KEY_ID,
+    // AWS_SECRET_ACCESS_KEY, etc. We don’t want to expose any of that. Only
+    // expose variables that are common between local Node.js and AWS Lambda.
+    //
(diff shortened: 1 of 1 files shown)


KSHITIZ6341:
Could we keep the env ordering canonical for every invocation, even when `activeEnvInvocations > 0`?

Before this change, `envs.sort(...)` always ran before computing `envHash`. With the sort inside the first-entry branch, an overlapping invocation skips sorting but still hashes `JSON.stringify(envs)`. If two concurrent requests carry the same env vars in different orders, they can now get different `envHash` values, which defeats the cache-key behavior described below and can unnecessarily re-import/evaluate modules.

One small fix would be to sort before the `if`, or sort a copied array and use that canonical array for both `process.env` population and the hash.

tsushanth:
Closing in favor of #493 by @jeshio which covers the same race condition and was opened first. Happy to contribute follow-up fixes once that lands.

Issue #49, opened 2025-02-24
convex-backend does not work on a raspberry pi 4


ArmNamGir:
I'm trying to install Convex on a Raspberry Pi 4 using Docker and a **docker-compose.yml** file.
Unfortunately, after building the images for the backend and the dashboard, I get the following error:
**dependency failed to start: container convex-backend-1 exited (132)**.

The convex-backend-1 container never starts, while convex-dashboard-1 runs correctly.

I tried rebuilding the images using the Dockerfiles, but the issue persists. Additionally, I'm unable to build Convex from source.

nipunn1313:
Hi! Can you get some more information from the log files? The logs from inside the container would be helpful.

Also regarding building convex from source - if you could provide any information (what you tried, what worked, what didn't work) - that would help debug.

mikecann:
Yes I failed to get it to work on the Pi 4b too unfortunately I cant remember what the issue was exactly. Something to do with low-level architecture.

I fortunately have a Pi 5 around too and managed to get it running on that. 

https://www.youtube.com/watch?v=cZMVQGJTNXo

It works just great, I use it every day :)

ArmNamGir:
Hi,

When I try to launch the convex-backend Docker container, it crashes instantly without generating any logs, neither in the container nor in Docker.

When building it from source, I encountered an error related to Rust, which fails to compile properly.

After some research, it seems that the issue comes from my Raspberry Pi, which is too old. As @mikecann mentioned, this could be due to the low-level architecture.

Since debugging information is very limited, it's difficult to pinpoint the exact cause of the problem.

So, I’ve decided to switch to a Raspberry Pi 5.

Thanks!

nipunn1313:
If you're able to get more logs - happy to take a look, but it's pretty hard to debug further otherwise. Good luck with the Raspberry Pi 5!

PR #503, opened 2026-07-11
fix: use portable arm64 codegen for self-hosted images

vadimmelnicuk:
## Summary

Use portable AArch64 code generation for self-hosted arm64 Docker images.

The repository-wide Linux arm64 configuration remains optimized for Neoverse N1. Only self-hosted Docker builds detected as `TARGETARCH=arm64` override this with `target-cpu=generic`.

This allows self-hosted backend images to run on baseline ARMv8-A CPUs such as the Raspberry Pi 4's Cortex-A72 without changing other arm64 builds or x64 images.

Addresses #49.

## Validation

- Built the complete self-hosted backend image natively on a Raspberry Pi 4 Model B 8GB using equivalent generic arm64 code generation.
- Started the backend image on the Pi and verified that it remained running rather than exiting with code 132.
- Verified that `/version` responded successfully.
- Verified the self-hosted dashboard running on the Pi.
- Ran `docker build --check` against the final scoped Dockerfile.
- With the repository Cargo configuration restored to Neoverse N1, verified on the Pi that:
  - BuildKit detected `TARGETARCH=arm64`.
  - Cargo received `target-cpu=generic`, not `target-cpu=neoverse-n1`.
  - A release-mode arm64 binary compiled and executed successfully.

The full backend validation used the same generic code-generation flags before they were scoped to the Docker build. The final Docker-only flag injection was then validated separately on the same Pi.

### DIFF
--- a/self-hosted/docker-build/Dockerfile.backend
+++ b/self-hosted/docker-build/Dockerfile.backend
@@ -20,6 +20,7 @@ RUN cargo chef prepare --recipe-path recipe.json --bin convex-local-backend
 # Uses cargo-chef's recipe from the previous stage for optimal caching
 FROM lukemathwalker/cargo-chef:latest-rust-1 AS build
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+ARG TARGETARCH
 
 # Update APT configuration for Docker builds
 # Sets retry attempts, assumes yes for prompts, and configures dpkg timeout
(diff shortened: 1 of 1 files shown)


gautamg795:
hey @vadimmelnicuk, thanks for the issue -- this was an oversight on our part. 
https://github.com/get-convex/convex-backend/commit/d6825caac38e9e1cd9a78b90a01bf52e4c3edc97 should fix this so there's no target-cpu enforced at all in the repo!

PR #506, opened 2026-07-17
fix(react): stabilize ConvexProviderWithAuth context value

dlabadini:
## Problem

`ConvexProviderWithAuth` builds a fresh object literal for `ConvexAuthContext.Provider`'s `value` on every render:

```tsx
<ConvexAuthContext.Provider
  value={{
    isLoading: isConvexAuthenticated === null,
    isAuthenticated,
    isRefreshing: isRefreshing && isAuthenticated,
  }}
>
```

`useConvexAuth()` reads this context with a plain `useContext()`, so every consumer re-renders whenever the value's *reference* changes - regardless of whether `isLoading`/`isAuthenticated`/`isRefreshing` actually changed. Since this object is rebuilt on every render of the provider, including renders triggered by parents above it that have nothing to do with auth, every render anywhere upstream of `ConvexProviderWithAuth` fans out to every `useConvexAuth()` consumer in the app.

## Fix

Wrap the value in `useMemo`, keyed on the three primitives it's built from (`isConvexAuthenticated`, `isAuthenticated`, `isRefreshing`). This is a pure reference-stabilization change:

- Doesn't change how often `ConvexProviderWithAuth` itself renders.
- Doesn't change any emitted value - only when a *new* value object is handed to consumers.
- Can't suppress a render that should happen, since all three source values are primitives read fresh each render and included in the dependency array.

## Testing

Verified locally against a real app: memoizing the value collapses redundant re-renders across `useConvexAuth()` consumers to only the renders where the derived auth state actually transitions, with no change in observable behavior.

---


Nicolapps:
Thanks for your PR! The fix is out in `convex@1.44.0`.

PR #509, opened 2026-07-23
Fix cursor when splitting later paginated query pages

### LINKED ISSUE #508: splitPaginatedQueryPage uses cursor: null for first split piece, causing duplicate results on non-first pages
## Bug

In `npm-packages/convex/src/browser/sync/paginated_query_client.ts`, the `splitPaginatedQueryPage` method always uses `cursor: null` for the first split piece:

```ts
// First split page: same cursor as original, but add endCursor at splitCursor
const firstSubscription = this.client.subscribe(
  paginatedQuery.canonicalizedUdfPath,
  {
    ...paginatedQuery.args,
    paginationOpts: {
      ...paginationOpts,
      cursor: null, // Start from beginning for first split
      endCursor: splitCursor,
    },
  },
);
```

`cursor: null` means "start from the very beginning of the table." This is only correct when splitting the first page of a paginated query. For any subsequent page (page 2, 3, ...), the first half of the split should start from the **original page's own start cursor**, not from the beginning.

## Reproduction

1. Use `usePaginatedQuery` with enough data that page 2+ is loaded (call `loadMore`).
2. Trigger a page split on a non-first page — either by growing that page beyond `initialNumItems * 2` or by the server returning `pageStatus: "SplitRequired"` / `splitCursor`.
3. Observe: the first split piece is subscribed with `cursor: null, endCursor: splitCursor`, so it returns all records from the very start of the table up to `splitCursor`, including items from earlier pages. Those items appear again in `results`.

## Root cause

`pageKeyToQuery` stores only `{ queryToken, unsubscribe }`. The start cursor used when the page was originally subscribed (via `addPageToPaginatedQuery`) is not retained anywhere, so `splitPaginatedQueryPage` cannot recover it and falls back to `null`.

## Fix

Store the page's start cursor alongside the subscription in `pageKeyToQuery`, then use it as the `cursor` for the first split piece:

```ts
// In addPageToPaginatedQuery, store startCursor:
paginatedQuery.pageKeyToQuery.set(pageKey, {
  ...subscription,
  startCursor: continueCursor,  // null for page 0, real cursor for later pages
});

// In splitPaginatedQueryPage, use stored startCursor instead of null:
paginationOpts: {
  ...paginationOpts,
  cursor: storedStartCursor,   // was hardcoded null
  endCursor: splitCursor,
},
```

Alternatively, parse the start cursor from the existing `queryToken` (a JSON string that includes `paginationOpts.cursor`), but storing it explicitly is cleaner.

shivanshgupta365:
## Summary

- retain the start cursor for each paginated query page
- reuse that cursor when splitting a later page
- preserve split cursors so recursively split pages keep correct boundaries
- add a regression test covering a split on page 2

Fixes #508

## Validation

- targeted Vitest regression test
- targeted ESLint and Prettier checks
- Rush build for the convex package

### DIFF
--- a/npm-packages/convex/src/browser/sync/paginated_query_client.test.ts
+++ b/npm-packages/convex/src/browser/sync/paginated_query_client.test.ts
@@ -1,0 +1,95 @@
+import { describe, expect, test } from "vitest";
+import { Value } from "../../values/index.js";
+import { BaseConvexClientInterface, Transition } from "./client.js";
+import { PaginatedQueryClient } from "./paginated_query_client.js";
+import { QueryToken } from "./udf_path_utils.js";
+import { Long } from "../../vendor/long.js";
+
+class TestBaseClient {
+  private transitionHandler: ((transition: Transition) => void) | undefined;
+  private nextToken = 0;
+  readonly subscriptions: Array<{
+    queryToken: QueryToken;
+    args: Record<string, Value>;
+  }> = [];
+  readonly results = new Map<QueryToken, Value>();
+
+  addOnTransitionHandler(handler: (transition: Transition) => void) {
+    this.transitionHandler = handler;
+    return () => {
+      this.transitionHandler = undefined;
+    };
+  }
+
+  subscribe(_name: string, args: Record<string, Value> = {}) {
+    const queryToken = `query-${this.nextToken++}` as QueryToken;
+    this.subscriptions.push({ queryToken, args });
+    return { queryToken, unsubscribe: () => {} };
+  }
+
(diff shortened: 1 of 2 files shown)


ianmacartney:
Is there anything here not covered by https://github.com/get-convex/convex-backend/commit/4aaae712235b1dd6534b2e5b803e4b4ea9dfea12 ?

shivanshgupta365:
No—the upstream commit covers the same cursor-preservation fix and includes a stronger result-level regression test. Thanks for landing it. I’m closing this PR as superseded.

PR #511, opened 2026-07-24
Add URL.canParse() static method to thefunction runtime

### LINKED ISSUE #510: Missing `URL.canParse()` in the Convex function runtime
The Convex function runtime's `URL` interface does not currently implement the [`canParse()`](https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static) static method, which causes Effect 4's [`urlFromString`](https://github.com/Effect-TS/effect/blob/b49284193f86737e411dc3dd19cfb1a8b9fa5d95/packages/effect/src/SchemaTransformation.ts#L1376) schema transformation to fail.

suhailopensource:
## What

Adds the standard [`URL.canParse(url, base)`](https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static) static method to the function runtime's `URL` class in `npm-packages/udf-runtime/src/00_url.ts`. The class previously implemented no statci methods so this Web API was missing.

## Why

Fixes #510. Libraries such as Effect 4 call `URL.canParse()` internally (e.g. its `urlFromString` schema transform) and break in Convex functions because the method is absent.

## How

`canParse` returns `true`/`false` without throwing implemented as the standard polyfill over the exisintng constructor:

static canParse(url: string | URL, base?: string | URL): boolean {
  try {
    new URL(url, base);
    return true;
  } catch {
    return false;
  }
}

This relies on the constructors existing behavior of throwing a TypeError on invalid input, which matches the WHATWG spec.

## Notes

- Additive only — no existing behavior changed, no new deps, no Rust changes.
- Formatted with \`dprint\` (no changes) and \`tsc --noEmit\` passes for \`udf-runtime\`.
- I couldnt find an existing runtimetest harness for \`URL\` (the code depends on \`performOp\`, which only runs inside the isolate). Ill be  like Happy to add an integration test wherever youd prefer.

---

### DIFF
--- a/npm-packages/udf-runtime/src/00_url.ts
+++ b/npm-packages/udf-runtime/src/00_url.ts
@@ -189,6 +189,15 @@ class URL {
   #urlInfo: UrlInfo;
   #searchParams: URLSearchParams;
 
+  static canParse(url: string | URL, base?: string | URL): boolean {
+    try {
+      new URL(url, base);
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
   constructor(url: string | URL, base?: string | URL) {
     let baseHref: string | null = null;
     if (base !== undefined) {


ianmacartney:
Thanks! I submitted a variant of this internally for review, also adding in URL.parse and tests

suhailopensource:
code merged

PR #518, opened 2026-07-30
Expose mutation commit timestamps in Convex clients

### LINKED ISSUE #182: Expose a mutation's commit timestamp (or requestId) on resolution, for state-library integrations built on BaseConvexClient
We're building a durable local-first layer on top of `BaseConvexClient` — persistent IndexedDB materialization of query results with a durable optimistic-command journal — which is exactly the use case the class documents ("Low-level client for directly integrating state management libraries with Convex").

The transition surface is excellent for the read side: `addOnTransitionHandler` gives us each query's new result together with `Transition.timestamp`, so we can stamp durable snapshots with the log position they're valid at. That timestamp is doing a lot of work for us: it's a result-version that covers the query's whole read set by construction.

The write side has one gap. To retire a **durable** optimistic overlay safely, we need to prove "this delivered query result includes my mutation's write" — i.e., `transition.timestamp >= my mutation's commit timestamp`. Both halves of that comparison already exist in the client:

- the wire's `MutationSuccess` carries `ts` (the commit timestamp), and
- `Transition.reflectedMutations` lists `{ requestId, result }` for mutations first reflected by that transition,

but neither is reachable from application code: `mutation()` resolves with only the function's return value, and the `requestId` it was assigned is private. So today there is no sound way to correlate our own mutation with the transition that first reflects it. (Matching by result value is not sound — values can be equal by coincidence, and ABA is real.)

**Ask:** any one of the following would fully unblock this, in our order of preference:

1. An opt-in on `mutation()` (e.g. a `MutationOptions` field) that resolves with `{ value, ts }` — the commit timestamp alongside the result; or
2. an option that surfaces the assigned `requestId` (so we can watch for it in `reflectedMutations`); or
3. a documented callback for "mutation reflected" keyed by something the caller can hold.

Option 1 is the smallest surface and composes directly with `getMaxObservedTimestamp()`/`Transition.timestamp`.

**Related small ask:** `ConvexReactClient` ships a `sync` getter returning its `BaseConvexClient` (the paginated query client is built on it), but it's omitted from the public type declarations. Blessing it in types would let React apps register transition handlers on the same query set as their hooks without a cast. If there's a reason it's untyped, guidance on the supported way to reach the underlying client from `ConvexReactClient` would serve the same purpose.

Happy to share more detail on the integration or test a preview build.

samvit:
## Summary

- add an opt-in `returnCommitTimestamp: true` mutation option that resolves with `{ value, ts }`
- preserve the existing return value and types for every mutation call that does not opt in
- expose the option through `BaseConvexClient`, `ConvexClient`, and `ConvexReactClient`
- make `ConvexReactClient.sync` and `BaseConvexClientInterface` public, including transition handlers and `getMaxObservedTimestamp()`

The returned `ts` is the mutation commit timestamp already present on `MutationResponse`. It is preserved through the request manager and only exposed after the client has transitioned through that timestamp, retaining the existing read-your-writes guarantee. Callers can use `result.ts.lessThanOrEqual(transition.timestamp)` to prove that a materialized query result includes the mutation.

This is useful for durable local-first/state-library integrations: they can retire persisted optimistic overlays using platform timestamps instead of adding application-specific command IDs to every query and coverage payload.

Closes get-convex/convex-js#182.

## Verification

- `npm run typecheck`
- `npm run build` (all CJS/ESM runtime and declaration targets)
- `npm run test-esm`
- public TypeScript API compile test covering opted-in and unchanged mutation return types, typed `ConvexReactClient.sync`, and `getMaxObservedTimestamp()`
- protocol-level runtime test covering:
  - commit timestamp propagation
  - no mutation resolution before the read-your-writes transition
  - unchanged default mutation, internal mutation, and action-restart behavior
- dependent `local-store` and `dashboard-common` builds
- Prettier and ESLint on all changed files

### DIFF
--- a/npm-packages/convex/src/browser/index.ts
+++ b/npm-packages/convex/src/browser/index.ts
@@ -20,11 +20,12 @@ export { BaseConvexClient } from "./sync/client.js";
 export type {
   BaseConvexClientOptions,
   MutationOptions,
+  MutationOptionsWithCommitTimestamp,
+  MutationResult,
   SubscribeOptions,
   ConnectionState,
   AuthTokenFetcher,
 } from "./sync/client.js";
-/** @internal */
 export type { BaseConvexClientInterface } from "./sync/client.js";
 export type { PaginationStatus } from "./sync/pagination.js";
 export type { ConvexClientOptions } from "./simple_client.js";
--- a/npm-packages/convex/src/browser/simple_client.ts
+++ b/npm-packages/convex/src/browser/simple_client.ts
@@ -12,6 +12,8 @@ import {
   BaseConvexClientOptions,
   ConnectionState,
   MutationOptions,
+  MutationOptionsWithCommitTimestamp,
+  MutationResult,
 } from "./sync/client.js";
 import {
   ExtendedTransition,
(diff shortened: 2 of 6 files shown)


ianmacartney:
We have a feature with this and more in 1.43 - you'll be able to store the commit timestamp in documents too!
In 1.43 you should be able to `return {ts: ctx.db.vars.commitTs}`

PR #522, opened 2026-08-06
Fix SQLite persistence materializing entire ranges in index_scan and load_documents (#495)

### LINKED ISSUE #495: Self-hosted SQLite backend: index_scan materializes the entire index range (ignores size_hint / no LIMIT), causing "too many system operations" + OOM on large tables
## Summary

On the self-hosted **SQLite** backend, `Persistence::index_scan` eagerly loads the **entire index interval into memory** before the query-layer `.take(N)` is applied. The SQL it runs has **no `LIMIT`** and the `size_hint` argument is **ignored**. As a result, an indexed read that should touch a handful of documents (e.g. `.withIndex(...).take(1)`) instead reads the whole range, which on a moderately large table fails with:

```
SystemTimeoutError: Your request timed out performing too many system operations.
```

On a small-RAM host it also drives the backend into memory pressure / OOM, because the full range is materialized into a `Vec`.

This is **not** version-specific — it reproduces on current `main` (verified against `f760918`).

## Root cause (with code pointers)

`crates/sqlite/src/lib.rs`:

- `index_scan(...)` takes `_size_hint: usize` — **prefixed with `_`, i.e. unused** (~L560).
- `_index_scan_inner(...)` (~L122) builds the SQL with `ORDER BY B.key {order}` and **no `LIMIT`**, then collects **every** matching row into `let mut triples = vec![]` before returning `Vec<...>`.

So the persistence layer returns the full interval; the `.take(N)` upstream only trims the already-materialized result. For a `by_created_at` / `by_x` index whose interval spans the whole table, that means "read the entire table" for a `take(1)`.

By contrast, the **Postgres** backend does the right thing — `crates/postgres/src/lib.rs` `index_scan` forwards `size_hint` and reads in paginated chunks with a real SQL `LIMIT` + cursor (`load_index_chunk`, `page_size`), so it does not have this behavior.

## Reproduction

1. Self-host the backend on the default SQLite persistence.
2. Populate an append-only table (e.g. a telemetry/event log) with a few hundred thousand rows.
3. From any query, run a tiny indexed read, e.g.:
   ```ts
   await ctx.db.query("events").withIndex("by_created_at").order("desc").take(1);
   ```
4. It fails with `SystemTimeoutError: ... too many system operations` (and spikes memory), even though only 1 document is requested. The same read is instant once the table is small again.

The same defect makes the snapshot **export** worker stall/blow up on a large table (it walks tables via the same index-scan path), and makes any retention sweep that reads a wide `createdAt < cutoff` range unable to ever catch up — the read it relies on hits the same wall, so the table grows unbounded.

## Impact

- Any indexed read (including the dashboard/admin reads and `convex export`) becomes unusable once a single table grows to ~hundreds of thousands of rows, regardless of the `.take()` limit.
- On constrained hosts the backend OOMs / livelocks rather than returning a clean error.

## Suggested fix

Make the SQLite `index_scan` stream lazily in bounded chunks (cursor + SQL `LIMIT`) and honor `size_hint`, mirroring the Postgres backend's `load_index_chunk` approach, instead of collecting the whole interval into a `Vec`.

Happy to put together a PR for the SQLite path along these lines if that direction sounds right — wanted to confirm the intended design first (whether SQLite is meant to stay a simple/dev-scale store) before investing in the change.

Iann29:
## Problem

On the self-hosted **SQLite** backend, both persistence read paths materialize entire ranges into memory instead of honoring their pagination parameters:

- `PersistenceReader::index_scan` ignores its `size_hint` argument (bound as `_size_hint`) and runs a query with **no `LIMIT`**, collecting every row in the interval into a `Vec` before the query layer's `.take(N)` is ever applied. An indexed read that should touch a handful of documents reads the whole range: on moderately large tables it fails with `SystemTimeoutError: Your request timed out performing too many system operations`, and on memory-constrained hosts it drives the backend into memory pressure / OOM (each row is also fully deserialized into a `ResolvedDocument` before anything is returned).
- `PersistenceReader::load_documents` ignores its `page_size` argument the same way and collects the entire documents-log range into a `Vec`. The boot-time search index bootstrap walks the whole log through this path, so the backend's startup working set scales with total log size.

Since #56013 the same is true of `load_documents_from_table`, which shares that code path and is now the one `DataSyncIterator` takes when it targets a single table.

This is the failure described in #495 and it reproduces on current `main`.

## Fix (two commits, one per read path)

Both fixes mirror the Postgres reader's design (`crates/postgres` already uses these parameters for batched cursor pagination):

**Commit 1 — `index_scan`:**
- `size_hint.clamp(1, 5000)` is the page size; each page is one query.
- The grouped subquery gains `ORDER BY key {order}` + `LIMIT {batch_size}`: since `GROUP BY index_id, key` runs over a prefix of the `(index_id, key, ts)` primary key, SQLite streams groups in key order and stops after one page.
- Each page resumes strictly after the last **scanned** key (`key > cursor` asc / `key < cursor` desc); pages re-query with `ts <= read_timestamp`, so pagination reads the same logical snapshot from the append-only log.
- **Tombstones count toward the page but are not emitted.** `B.deleted` moved from the join filter into the projection: with the old join filter, a page consisting entirely of tombstoned keys would be indistinguishable from the end of the interval and a "fewer rows than requested ⇒ done" termination would silently drop the rest of the range.
- The retention snapshot is validated after each page is read and before any of its rows are yielded — the same contract as the Postgres reader.
- Also fixes a latent row-mapper failure path: tombstoned index rows carry `document_id = NULL`, which the old `row.get::<_, Vec<u8>>(2)` would have failed on (it never saw them only because the join filtered tombstones out).

**Commit 2 — `load_documents` and `load_documents_from_table`:**
- `page_size` is now honored (guarded with `.max(1)`); each page is one query with `LIMIT`, resuming strictly after the `(ts, table_id, id)` primary key of the last row read (composite strict bound, mirrored for `Desc`).
- Deletes are legitimate log entries here, so nothing is filtered after the query runs and short-page termination is safe.
- Retention is validated per page via `validate_document_snapshot`, and the stream keeps the existing `.cooperative()` wrapping.
- `load_documents_from_table` shares the paginated path, taking the tablet restriction as one more bound on the same query. Restricting in SQL — rather than filtering rows out after they are read — is what keeps short-page termination correct: `LIMIT` counts only rows that will be yielded. A filter applied after the limit would hand back short pages on an interleaved log and end the stream early.
- Placeholder numbers are derived from the parameter vector as each one is bound, so adding the tablet parameter cannot silently renumber the cursor's.

Error surfacing is now lazy (first poll of the stream) rather than eager at call time, matching the Postgres reader.

## Rebased onto current `main`

#56013 landed after this PR was opened: it made `PersistenceReader::load_documents_from_table` a required trait method (dropping the default body that the SQLite reader relied on) and pushed the tablet filter down into SQL for it. That optimization is preserved here — the filter stays in the query — and pagination is added underneath it, so the single-table `DataSyncIterator` path gets both. The free `load_docs` helper is gone; the query is built inline where the cursor and tablet bounds are assembled.

## Measurements

We hit this in production self-hosting and measured the fix against a **snapshot of a real production database** (2.3 GB SQLite, ~1.5 M documents-log rows, ~3.2 M index rows): same machine, same base commit, network-isolated boot, RSS sampled every 5 s. These numbers were taken before the rebase; the pagination design is unchanged.

| Backend | Boot RSS behavior |
|---|---|
| Unpatched | Climbs without bound: **6.8 GB at 7 min 25 s**, killed by a safety cap while still climbing (in production this plateaued at 14–16 GB and pushed the host into reclaim thrashing) |
| Commit 1 only | Climbs to ~5.1 GB in 95 s, then holds a stable plateau at **~5.3 GB** (the remaining climb is `load_documents` during search index bootstrap) |
| Commits 1 + 2 | **~355 MB**, flat for the entire observation window |

## Tests

`crates/sqlite` had no test module; this PR adds one (most of the diff — the production change is small):

- `paginated_scan_matches_reference` — 25 keys with updates, tombstones landing on page boundaries, and writes newer than the read snapshot, checked against an in-memory reference model across 2 read timestamps × 2 intervals × `Asc`/`Desc` × size hints `{1, 2, 3, 10_000}`.
- `page_of_tombstones_does_not_terminate_scan` — 10 consecutive tombstoned keys forming whole pages with no live rows, the case a naive termination check breaks on.
- `load_documents` equivalents: full-log reference comparison across page sizes/orders/partial ts ranges, plus a run of many documents written at a **single timestamp** with pages breaking inside it (the `(table_id, id)` cursor tiebreak is the fragile spot), and empty-table/empty-range cases.
- `paginated_load_from_table_matches_reference` — two tablets interleaved at every timestamp, compared against a filtered reference across both orders and page sizes `{1, 3, 5, 10_000}`. This is the test that fails if the tablet restriction moves out of SQL, or if a placeholder number is hardcoded once a second parameter exists.

Validation: `cargo test -p sqlite` green (7 tests), `cargo fmt` clean, `cargo clippy -p sqlite --lib --tests` introduces no new warnings. Each commit builds and passes on its own.

Fixes #495

### DIFF
--- a/crates/sqlite/src/lib.rs
+++ b/crates/sqlite/src/lib.rs
@@ -43,7 +43,6 @@ use common::{
     },
     query::Order,
     runtime::CoopStreamExt as _,
-    try_anyhow,
     types::{
         IndexId,
         PersistenceVersion,
@@ -55,10 +54,7 @@ use common::{
         TabletId,
     },
 };
-use futures::{
-    stream,
-    StreamExt,
-};
+use futures::StreamExt;
 use futures_async_stream::try_stream;
 use parking_lot::Mutex;
 use rusqlite::{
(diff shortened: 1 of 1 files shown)


Iann29:
Pushed a second commit extending the fix to `load_documents`, which has the same unbounded materialization (ignored `page_size`, no `LIMIT`). Measuring against a snapshot of our production database (2.3 GB SQLite) showed it was actually the dominant term at boot: the search index bootstrap walks the whole documents log through that path.

Boot RSS on identical data, same base commit, network-isolated, sampled every 5 s:

- unpatched: 6.8 GB after 7.5 min and still climbing (killed by a safety cap; in production this plateaus at 14–16 GB)
- `index_scan` fix only: stable plateau at ~5.3 GB
- both fixes: **~355 MB, flat**

PR body updated with the full methodology. Happy to split into two PRs if you'd prefer smaller reviews.

diogorocha18:
I'm having the same problem.

Andrew-Sem:
Nice work — we ran into the same SQLite memory issue in production. I opened #539 for a smaller part of the startup path, `max_ts()`. This PR bounds the general document-loading path; #539 avoids it for this lookup by using `SELECT MAX(ts)`, so no documents need to be decoded. Linking the two since the fixes complement each other.

Issue #523, opened 2026-08-07
@convex-dev/workos: fetchAccessToken reports transient refresh failures as sign-outs and ignores forceRefreshToken


jmosbacher:
## Summary

`ConvexProviderWithAuthKit`'s auth adapter (`npm-packages/@convex-dev/workos/src/index.tsx`, current on main and in 0.0.3) has two defects that combine into a frequent, sticky false "signed out" state in apps using WorkOS AuthKit:

```ts
const fetchAccessToken = useCallback(async () => {
  try {
    const token = await getAccessToken();
    return token;
  } catch {
    return null;
  }
}, [getAccessToken]);
```

### 1. Every `getAccessToken` failure is mapped to `null` (= "user signed out")

`AuthenticationManager` treats a `null` from `fetchToken` as the user having signed out (`setAndReportAuthFailed`), flipping `useConvexAuth().isAuthenticated` to `false`. But authkit-js `getAccessToken()` also throws for **transient** conditions: network failures, refresh-lock acquisition timeouts, and (since authkit-js 0.20.2, which added `RefreshError.isTransient` / `RefreshTimeoutError`) HTTP 408/429/5xx refresh responses — all rethrown to the caller while the WorkOS session remains alive.

This bites constantly in practice because authkit-js deliberately pauses its auto-refresh while `document.hidden`, and its access token (5-minute default) lives in memory. So the token is expired at nearly every tab refocus, forcing a refresh exactly at the moment of window/network churn — where transient failures are most likely.

### 2. `forceRefreshToken` is ignored

The custom-auth docs state the contract: *"Make sure to fetch a new token when `forceRefreshToken` is true."* This adapter never passes it, so when the backend rejects a token that authkit-js still considers valid client-side, the adapter re-serves the identical token and `tryToReauthenticate` reports auth failure. `ConvexProviderWithClerk` gets this right (`getToken({ skipCache: forceRefreshToken })`).

### Aggravator: the failure is sticky

After `setAndReportAuthFailed`, `ConvexProviderWithAuth` only re-runs `client.setAuth` when the `fetchAccessToken` identity (or auth-provider state) changes. A failed refresh changes no AuthKit React state — `user` stays set — so nothing ever retries. The app shows its signed-out UI indefinitely while the WorkOS session is alive; clicking "sign in" round-trips through the hosted flow silently and "fixes" it, which masks the bug as an aggressive-logout complaint.

## Repro sketch

[...]

PR #529, opened 2026-08-12
Fix WorkOS AuthKit token refresh handling

jokrasno:
## Summary

- Honor Convex's `forceRefreshToken` through AuthKit's `forceRefresh` option, preserving the zero-argument cached-token flow otherwise.
- Return `null` immediately when AuthKit requires login.
- Retry other token-fetch errors up to three total attempts, waiting 250 ms and 500 ms. Persistent failures resolve to `null`; no token-fetch rejection escapes to Convex's auth manager.
- Update AuthKit React to the patch release with transient refresh-error semantics.

Related to #523. This handles short-lived failures but does not provide automatic recovery after the retry budget is exhausted. A prolonged outage can still report unauthenticated until auth is reconfigured; this PR does not claim to fully solve that case.

## Follow-up validation

- Seven adapter tests pass, covering terminal login failures, transient recovery in cached/forced modes, bounded persistent failures, and login-required during retry.
- TypeScript checking passes for the adapter and tests.
- Tests used byte-identical source copies in a small isolated environment with Convex 1.42.3 and AuthKit React 0.16.2, one Vitest worker.
- Formatting and `git diff --check` pass.
- Full monorepo lint/build and browser integration were not rerun for this follow-up.

### DIFF
--- a/npm-packages/@convex-dev/workos/package.json
+++ b/npm-packages/@convex-dev/workos/package.json
@@ -19,7 +19,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@workos-inc/authkit-react": "^0.16.0"
+    "@workos-inc/authkit-react": "^0.16.2"
   },
   "devDependencies": {
     "@types/react": "^18.0.0",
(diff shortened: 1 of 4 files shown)


erquhart:
@cubic-dev-ai review

erquhart:
@greptileai review

jokrasno:
Good catch, Convex doesn't catch those rejections. Added bounded retries and a null fallback instead of throwing. All 7 adapter tests pass. Long outages still need a separate recovery path; noted that in the description.

erquhart:
Thanks @jokrasno, re-reviewing this now

Issue #525, opened 2026-08-09
Self-hosted: in-process searchlight disk cache and segment LRUs are uncoordinated — RAM growth from deleted-but-mapped segments, meta.json watcher log flood, and ENOENT query failures


Sjotie:
## Environment

- Hosting: Railway (self-hosted), single instance, persistent volume at `/convex/data`
- Image: `ghcr.io/get-convex/convex-backend:latest` (currently `sha256:240a2e9115dab3c13145b96532d1cc8bc63c5e2d112d95f80e592a75242bfa75`)
- Resources: 32 GB cgroup memory limit
- Dataset: ~21 GB of text + vector index segments (one large vector index with ~650k chunks/embeddings, several smaller text + vector indexes)

## Summary

In self-hosted deployments, search runs through `InProcessSearcher`, which layers two caches that are not coordinated with each other:

1. **An archive disk cache** that extracts segment archives into `tmpdir/<uuid>/` and evicts by deleting extracted directories — hardcoded to 500 MiB with no env knob ([`crates/search/src/searcher/in_process.rs#L147`](https://github.com/get-convex/convex-backend/blob/main/crates/search/src/searcher/in_process.rs#L147)).
2. **Two in-memory segment LRUs** (`TextSegmentCache`, `VectorSegmentCache`) that hold live `tantivy::Searcher`s and qdrant `Segment`s (each with its own RocksDB instance), **keyed by extracted path** and **counted per entry** (`SizedValue::size() = 1`, defaults 120 + 120 — [`segment_cache.rs`](https://github.com/get-convex/convex-backend/blob/main/crates/search/src/searcher/segment_cache.rs), [`searchlight_knobs.rs#L82-L92`](https://github.com/get-convex/convex-backend/blob/main/crates/search/src/searcher/searchlight_knobs.rs#L82-L92)).

Because the disk cache deletes directories that the memory LRUs still reference, any deployment whose working set exceeds 500 MiB of extracted segments (i.e. any non-trivial dataset) ends up in one of two failure modes depending on which side loses the race:

- **Memory-LRU entries outlive their directories** → RAM grows to the cgroup limit via deleted-but-still-mapped files, plus a tantivy file-watcher log flood.
[...]

PR #531, opened 2026-08-19
Stop cached search index readers from polling deleted meta.json files, and make the archive cache ceiling configurable

Iann29:
## Problem

#525 describes an in-process searcher whose archive disk cache and segment LRUs are uncoordinated, and lists three consequences: RAM held by deleted-but-mapped segments, a `meta.json` watcher log flood, and ENOENT query failures. This PR does not implement the coordination fix (suggestion 1 in that issue); it takes suggestions 2 and 3, which stand on their own, and adds two cleanup bugs found while reading that code.

We hit the same thing on an unrelated self-hosted deployment and arrived at the same two mechanisms independently, which is why they are proposed separately: neither depends on the larger refcounting change, and both are useful whether or not it lands.

## The watcher flood is unbounded in time, not just in volume

`index_reader_for_directory` is only ever pointed at a directory the archive cache extracted: an immutable snapshot of one already-committed segment. Both call sites assert `segment_readers().len() == 1`, take `reader.searcher()` immediately, and never reload. Nothing in the process commits to those directories, so the default `ReloadPolicy::OnCommit` buys nothing and costs a thread per reader polling `meta.json` every 500 ms.

The poll outlives the file. Once the archive cache deletes the extracted directory, the watcher cannot stop or succeed: `compute_checksum` logs `Failed to open meta file` and the thread sleeps and retries forever. On our instance that was a sustained ~137 lines/sec, several GB of JSON logs per day, from watchers that can never recover. #525 reports 117 lines/sec from the same mechanism.

`ReloadPolicy::Manual` registers no watch callback at all, so no thread is spawned.

## Two cleanup bugs in the archive cache

Both let extracted directories outlive the entry that owns them, and neither is mentioned in #525.

**`IndexTempDir` never had the `Drop` its construction site documents.** The comment at the call site reads "Create this right away so its Drop impl (which deletes the path) runs even on failure", but only `IndexTempDirWithSize` has one, and a fetch reaches that type only after it knows the extracted size. A fetch that fails or times out after `extract_archive` created the directory therefore drops the handle without deleting anything, leaking a directory that nothing will ever account for or clean up. This PR gives `IndexTempDir` the `Drop` it was documented to have, disarmed when `IndexTempDirWithSize` takes over so the transfer does not schedule a deletion of a live entry.

**The cleanup thread clears the readonly bit on the wrong directory.** It clears it on the entry's top-level directory, but the `FragmentedVectorSegment` path extracts into a nested `<uuid>/segment` and marks *that* readonly. Removing the tree then has to unlink files inside a directory with no write bit, which fails with `EACCES` for any process not running as root — and that error hits the `panic!` arm. Since the release profile sets `panic = "abort"`, that is a process abort on the first such eviction, not a silently dead cleanup thread. It is masked today only because the published image runs as root. This PR clears the bit recursively; the `panic!` arm is left alone, so any other IO error still behaves as before.

## Configurable ceiling

`SEARCHLIGHT_DISK_CACHE_MIB` (default 500, unchanged behavior) joins the other `env_config` knobs in `searchlight_knobs.rs`. As #525 notes, this was the only piece of the searcher without one. `SearcherImpl::new` already logs the effective ceiling at startup, so the configured value is visible without a metrics endpoint.

## Observability

`SEARCHLIGHT_USED_BYTES` already tracks the cache's disk usage, but a self-hosted backend exports no Prometheus endpoint, so on those deployments there is no way to tell a working cache from one that is thrashing. This PR logs one line per directory the cleaner actually removes, with its size and the configured ceiling. The rate is bounded by evictions, not by cache hits — deliberately not the per-fetch "space used" accounting next to it, which stays at `debug` precisely because log volume is the failure mode being addressed.

## Tests

`crates/search` had no test module; this PR adds one.

- `index_reader_for_directory_spawns_no_meta_file_watcher` — asserts no watcher thread exists after opening a reader, with a **positive control** that first asserts tantivy's default policy *does* spawn one for the same fixture, so a zero cannot mean the fixture was wrong. Linux-only: it reads thread names from `/proc`.
- `nested_readonly_directories_can_be_removed` — builds the `<uuid>/segment` layout with the readonly bit where `extract_segment` puts it, asserts the old sequence fails with `PermissionDenied`, then that the recursive clear succeeds. Running as root it self-skips (there is nothing to regress against) and says so on stderr.
- `index_temp_dir_deletes_its_directory_exactly_once` — the failure path, the transfer to `IndexTempDirWithSize` (which must *not* delete a live directory), and eviction.
- `evicted_entries_are_deleted_from_disk` — drives real evictions through `ArchiveCacheManager`: six archives into a cache sized for two, asserting the extracted directories leave the disk. Under `RUST_LOG=search=info` it also shows the new log line firing exactly four times, once per eviction and never per fetch.

Each fix was checked by reverting it with its test in place; each corresponding test fails. Validation: `cargo test -p search` green, `cargo fmt` clean, `cargo clippy -p search --lib --tests` introduces no new warnings. Each commit builds and passes on its own.

`cargo clippy -p search --all-targets` currently fails on `benches/memory_index.rs` (`unresolved import value::assert_obj`) on `main`, independently of this PR.

## What this does not do

It does not coordinate the two caches, which is the root cause in #525 and the only thing that removes the race. With this PR the operator can raise the disk ceiling until the working set fits, which is what the issue reporter achieved by tuning the LRU knobs — still tuning around the race rather than removing it. The log flood, however, is gone outright: with no watcher there is nothing left to flood.

We have not yet run a patched build in production, so there is no before/after RSS measurement here — the evidence is the tests above plus the production symptom measurements described in #525 and repeated on our own instance.

Addresses #525.

### DIFF
--- a/crates/search/src/archive/cache.rs
+++ b/crates/search/src/archive/cache.rs
@@ -349,7 +349,7 @@ impl<RT: Runtime> ArchiveCacheManager<RT> {
         max_concurrent_fetches: usize,
         rt: RT,
     ) -> anyhow::Result<Self> {
-        let cleaner = CacheCleaner::new(rt.clone());
+        let cleaner = CacheCleaner::new(rt.clone(), max_size);
         let cache = AsyncLru::new(rt.clone(), max_size, max_concurrent_fetches, 200, "cache");
         let this = Self {
             path: local_storage_path.as_ref().to_owned(),
@@ -521,10 +521,11 @@ struct CacheCleaner {
 }
 
 impl CacheCleaner {
-    fn new<RT: Runtime>(rt: RT) -> Self {
+    fn new<RT: Runtime>(rt: RT, max_size: u64) -> Self {
         let (cleanup_tx, cleanup_rx) = mpsc::unbounded_channel();
-        let cleanup_handle =
-            Arc::new(rt.spawn_thread("search_cache_cleaner", || cleanup_thread(cleanup_rx)));
+        let cleanup_handle = Arc::new(rt.spawn_thread("search_cache_cleaner", move || {
+            cleanup_thread(cleanup_rx, max_size)
+        }));
         Self {
             cleanup_tx,
             _cleanup_handle: cleanup_handle,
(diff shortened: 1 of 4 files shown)


smoussa:
Also have this issue. Would love to get this merged in.

emmaling27:
Thanks for the contribution @Iann29! I broke this into a couple of commits and merged them in, most importantly bf09681d31b19c812cec96986b95a724ab20e501, 1a892236adb3dac260fbe742cd1a835c4e448f42, c020d72ab084cac98dd578497b86e7c9c9e97ec7, a16dada1d073e5a5c5eaabd25cbe7f2a4c4f18f9

Issue #317, opened 2026-01-14
VectorCompactor crashes with 'No such file or directory (os error 2)' on Railway self-hosted deployment


ashot:
## Environment
- Hosting: Railway (self-hosted)
- Image: `ghcr.io/get-convex/convex-backend:latest`
- Resources: 32 vCPU / 32 GB RAM (maxed out)
- Volume: 50 GB persistent volume mounted at `/convex/data`
- Region: US West (California)

## Problem
When enabling a vector index on a table with ~77k+ embeddings (3072 dimensions), the VectorCompactor crashes repeatedly with:

```
ERROR common::errors: VectorCompactor died: No such file or directory (os error 2)
ERROR database::search_index_workers::retriable_worker: VectorCompactor died, num_failures: 2. 
Backing off for 93232ms (max: 600000ms), expected: false
```

The crash loop eventually causes 502 Bad Gateway errors across the entire Convex backend, making the service unusable.

## Schema
```typescript
.vectorIndex('by_embedding', {
  vectorField: 'embedding',
  dimensions: 3072,
  filterFields: ['source', 'direction', 'category'],
})
```

## Steps to Reproduce
1. Deploy self-hosted Convex on Railway with persistent volume at `/convex/data`
2. Create a table with ~77k documents containing 3072-dimension embeddings
3. Add a vector index to the schema
4. Deploy - the vector index backfill starts
5. VectorCompactor crashes with "No such file or directory" error
6. Backend becomes unstable with 502 errors

## Expected Behavior
Vector index backfill should complete successfully without crashing.

## Workaround
Currently keeping the vector index disabled to maintain service stability.

## Hypothesis
The VectorCompactor may be writing temporary files to a non-persisted location (like `/tmp`) instead of the mounted volume at `/convex/data`. When the process restarts or the container cycles, these temp files are lost, causing the "No such file or directory" error.

## Questions
1. Are there any known issues with VectorCompactor on self-hosted deployments?
2. Is there a way to configure where VectorCompactor stores its temp files?
3. Are there minimum resource requirements for vector index operations with large embeddings?

emmaling27:
Should be fixed in bf09681d31b19c812cec96986b95a724ab20e501

Issue #192, opened 2025-08-19
No such file or directory (os error 2)


xixixao:
When node is not on path, the backend now fails during deploy (before, like 6 months ago, I don't think Node was required during deploy).

1. Is node needed for all deploys?
2. Can we add better error handling?

For 2, it would probably suffice to change:

```rust
async fn check_node_version(node_path: &str) -> anyhow::Result<()> {
        let cmd = TokioCommand::new(node_path)
            .arg("--version")
            .output()
            .await?;
```
to
```rust
async fn check_node_version(node_path: &str) -> anyhow::Result<()> {
        let cmd = TokioCommand::new(node_path)
            .arg("--version")
            .output()
            .await
            .context("`node` is not on PATH")?;
```

Current error (quite cryptic):
```
Hit an error while pushing:                                                                                                             │
│ Failed to initialize inner local node executor                                                                                                                                            │
│                                                                                                                                                                                           │
│ Caused by:                                                                                                                                                                                │
│     0: Failed to initialize inner local node executor                                                                                                                                     │
│     1: No such file or directory (os error 2)                                                                                                                                             │
│                                                                                                                                                                                           │
│ Stack backtrace:                                                                                                                                                                          │
│    0: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from                                                                                                                │
│    1: <async_once_cell::InitFuture<T,F> as core::future::future::Future>::poll::{{closure}}                                                                                               │
│    2: <async_once_cell::InitFuture<T,F> as core::future::future::Future>::poll
```

nipunn1313:
Nice Let's definitely do (2).

I believe node is required if there are any "use node" actions in the project.

xixixao:
@nipunn1313 but as far as I can tell, it's required on startup of the backend, even a completely empty one, that cannot know whether there are any node actions.

PR #532, opened 2026-08-21
search: make the in-process searcher's archive cache size configurable

Sjotie:
`InProcessSearcher::new` hardcodes the on-disk archive cache at `bytesize::mib(500u64)`. That cache holds the *unpacked* text and vector segments, and it is not coordinated with the in-memory segment LRUs (`MAX_TEXT_LRU_ENTRIES` / `MAX_VECTOR_LRU_ENTRIES`), which key on path and are sized by entry count rather than bytes.

Once a deployment's unpacked working set passes 500 MiB, the two caches disagree: the disk cache evicts a segment directory while an LRU still holds that segment open. Re-opening it then fails. Symptoms, in the order they usually get noticed:

- intermittent `No such file or directory (os error 2)` surfacing out of `vectorSearch`, retried successfully on the next attempt because the segment is re-fetched under a fresh path
- a tantivy `meta.json` file-watcher log flood from watchers polling directories that no longer exist
- memory growth from deleted-but-mapped segment files, which stay resident until the LRU entry is finally evicted

This affects self-hosted backends, where `InProcessSearcher` is the searcher.

### Evidence

A/B on identical data (40k chunks, ~490 MiB of segments, both LRUs capped at 4 so the disk cache is the only variable), same binary with only this knob changed:

| archive cache | searches OK | ENOENT |
|---|---|---|
| 100 MiB (below working set) | 0 / 250 | 289 |
| 2048 MiB (above working set) | 400 / 400 | 0 |

At 100 MiB the backend additionally aborted after five runs (exit 133), which is the failure mode when a mapped segment disappears underneath a live searcher.

### The change

Expose the size as `MAX_ARCHIVE_CACHE_SIZE_MIB` next to the other searchlight knobs, defaulting to 500 so behaviour is identical unless the knob is set. Two values are rejected at construction rather than silently accepted:

- **zero** — a cache that evicts every segment immediately after fetch reintroduces exactly the race the knob exists to avoid
- **sizes whose byte count overflows `u64`** — the release profile aborts on arithmetic overflow, so this would be a panic rather than an error

The resolution lives in a small free function with unit tests for the default, the zero case, and the overflow case.

### Notes

This is the smallest change that makes the problem addressable by operators; it does not fix the underlying coherence issue between the disk cache and the segment LRUs, which is described in #525. #317 and #192 report the same ENOENT and are, I believe, the same root cause.

Formatted with the repo's pinned nightly `rustfmt`.

### DIFF
--- a/crates/search/src/searcher/in_process.rs
+++ b/crates/search/src/searcher/in_process.rs
@@ -4,6 +4,7 @@ use std::collections::BTreeMap;
 /// - InProcessSearcher implementation
 use std::sync::Arc;
 
+use anyhow::Context as _;
 use async_trait::async_trait;
 use common::{
     bootstrap_model::index::{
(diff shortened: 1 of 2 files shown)


emmaling27:
This is already done

PR #533, opened 2026-08-21
storage: make LocalDirStorage::delete_object idempotent

Sjotie:
`LocalDirStorage::delete_object` calls `fs::remove_file` and propagates its error, so deleting an object that is already gone returns `ENOENT` instead of succeeding. Every other `Storage` implementation here is backed by S3 semantics, where `DeleteObject` on a missing key is a success — "delete" means "ensure it is gone", not "remove exactly once".

The practical effect on a self-hosted backend is that `SystemTableCleanupWorker` dies whenever it encounters an `_exports` row whose blob is no longer on disk. Because that worker deletes from several system tables in sequence, one unremovable object stops the rest of the cleanup as well: `_exports` blobs, source packages and `_session_requests` all stop being reclaimed, and the retry loop reproduces the same crash on the next tick indefinitely. On one deployment this ran at roughly 95 failures/day for five weeks, and the wedge is in the data rather than the process, so restarts do not clear it.

### The change

Treat a missing object as success — but only a missing object. If the storage root itself has vanished (volume unmounted, directory wiped), the error still surfaces, because in that case the caller would otherwise commit a deletion that never durably happened. That distinction is the reason this is a few lines rather than one.

The body is factored into a free function so the semantics can be unit-tested without constructing a `Runtime`. Five tests cover: removing an existing file, deleting twice, deleting something that never existed, a vanished storage root (must error), and a directory at the path (must keep erroring — only `NotFound` is tolerated).

### Notes

Root-cause write-up for the wider issue is in #525; this half of it stands on its own.

Formatted with the repo's pinned nightly `rustfmt`.

### DIFF
--- a/crates/storage/src/lib.rs
+++ b/crates/storage/src/lib.rs
@@ -1222,14 +1222,10 @@ impl Display for StorageUseCase {
     }
 }
 
-/// Core of `LocalDirStorage`'s [`Storage::delete_object`], kept as a free
-/// function so the deletion semantics are unit-testable.
-///
-/// Deleting an already-absent object succeeds, matching S3 DeleteObject
-/// semantics: delete means "ensure it is gone". Only a missing object is
-/// tolerated — if the storage root itself is gone (unmounted or wiped),
-/// report the error rather than letting the caller commit a delete that
-/// never durably happened.
+/// Body of `LocalDirStorage::delete_object`, kept as a free function so the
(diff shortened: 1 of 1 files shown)


emmaling27:
Thanks for the contribution! merged in a16dada1d073e5a5c5eaabd25cbe7f2a4c4f18f9

PR #537, opened 2026-08-27
Fix Windows local executor lifecycle

jviningsli:
## Summary

- Keep current Windows local backends alive long enough to observe CLI stdin EOF and complete application shutdown, while keeping the backend window hidden.
- Assign each Windows Node action executor to a kill-on-close Job Object so abrupt backend termination cannot strand the executor process tree.
- Make executor shutdown explicit: stop and reap Node, close the named-pipe client, remove the executor temp directory with a bounded Windows retry, and propagate persistent cleanup failures.
- Preserve the existing SIGTERM fallback for older backend versions and document activation, verification, and rollback.

## Failure reproduced

Force-terminating a running Convex CLI previously ended its Node-spawned backend before Rust cleanup could run. The backend's Node action executor, named pipe, and temporary source directory could therefore outlive the dev session. The implementation separates the backend from the CLI's Windows process group but retains the stdin ownership pipe; the backend now receives EOF, drains, and cleans up. The Job Object covers abrupt backend termination.

## Verification

- cargo test -p node_executor --lib -- --nocapture (7 passed)
- cargo build -p local_backend --bin convex-local-backend
- Convex package Prettier, ESLint, and TypeScript checks
- Built the CLI and backend, ran a real Node action, and verified ordinary CLI exit left no backend PID, executor PID, named pipe, or executor temp directory.
- Repeated with forced CLI termination and verified all four resources were absent after shutdown.

<a href="https://cubic.dev/pr/get-convex/convex-backend/pull/537?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -48,6 +48,48 @@ Build and run the local backend from the source in this repo:
 just run-local-backend
 ```
 
+#### Local process lifecycle on Windows
+
+The Convex CLI owns every local backend it starts. Current CLI and backend
+versions keep the backend's standard input connected; closing that pipe asks the
+backend to drain requests, stop workers, and remove its temporary files. The
+pipe also closes when the CLI exits unexpectedly. On Windows, the CLI starts
+current backends in a hidden, detached process group. This prevents Windows from
+ending the backend with its Node parent before the backend can observe the
+closed pipe.
+
+The local backend owns each Node action executor. On Windows it assigns the
+executor to a Job Object configured with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.
+Windows therefore terminates the executor if the backend is terminated before
+Rust cleanup can run. During an ordinary shutdown, the backend stops and reaps
+the executor before removing its temporary directory. Removal uses a short,
+fixed retry window for transient Windows sharing violations and reports a
+persistent cleanup failure instead of silently leaking the directory.
+
+Run the focused Windows lifecycle tests after changing this boundary:
+
+```powershell
+cargo test -p node_executor force_killing_job_owner_terminates_assigned_process
+cargo test -p node_executor ordinary_drop_stops_executor_and_removes_temp_dir
+cargo test -p node_executor explicit_shutdown_stops_executor_and_removes_temp_dir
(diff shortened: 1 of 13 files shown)


jviningsli:
Follow-up candidate f13de4bee5af856dcf709bd28215329305d6f45c resolves all six findings from the Cubic review of 452b14a6708c7424f88c4c17e8681815da67d0f0. It now spawns Windows Node executors suspended and assigns/resumes them inside the Job Object; keeps the stdin watcher off Tokio runtime-owned blocking workers; routes startup, Drop, and explicit shutdown through bounded process-first cleanup; avoids unbounded child waits on Tokio workers; probes the selected backend binary before adding --shutdown-on-stdin-close; and corrects rollback guidance to require a matching CLI/backend pair. Validation on Windows: file-local rustfmt completed, git diff --check passed, and an isolated Cargo harness compiled the repository windows_job.rs and passed force_killing_job_owner_terminates_assigned_process. A full workspace cargo check could not reach candidate compilation on this machine because the existing aws-lc-sys dependency requires unavailable CMake/NASM tooling; upstream automation remains the complete build gate.

jviningsli:
Candidate 15777863ae09209508138f2c768758896b0bfcaa addresses the three review findings on f13de4bee5af856dcf709bd28215329305d6f45c. The CLI now validates the binary with --help, then probes the intentionally hidden lifecycle flag together with --version, so current backends enable stdin ownership while older backends reject the probe without starting. Executor cleanup jobs are registered with a single owner, joined during explicit shutdown, and joined again when that owner drops, so fatal runtime teardown cannot outrun them. The shared cleanup routine releases failed process ownership, always attempts bounded temporary-directory cleanup, and combines both errors. A focused Windows test now proves pending cleanup is joined before shutdown returns. File-local rustfmt and git diff --check pass; the previously reported full-workspace aws-lc-sys toolchain limitation remains unchanged.

PR #539, opened 2026-08-28
Avoid loading the document log in SQLite max_ts

Andrew-Sem:
## Summary

- override SQLite `max_ts()` with `SELECT MAX(ts)` instead of the generic document stream
- preserve `MaxRepeatableTimestamp` and the exclusive `Timestamp::MAX` upper bound
- add focused tests for empty persistence, globals, tombstones, malformed values, and timestamp boundaries

## Problem

The generic `PersistenceReader::max_ts()` asks `load_documents(..., Desc, 1)` for the newest document. SQLite currently collects and decodes the complete matching document log before returning that stream, so a one-row lookup can use memory proportional to the full history during backend startup.

This implementation reads only the aggregate timestamp. It intentionally no longer performs the incidental full-row validation caused by the generic eager scan; `max_ts()` only promises the largest persisted timestamp.

PR #522 would also bound the generic `load_documents()` path. This narrower override remains useful independently because it avoids document decoding altogether and keeps the timestamp lookup explicit.

Related: #495, #522

## Validation

- `mise exec -- cargo fmt -p sqlite -- --check`
- `mise exec -- cargo test -p sqlite`
- `mise exec -- cargo clippy -p sqlite --all-targets -- -D warnings`

### DIFF
--- a/crates/sqlite/src/lib.rs
+++ b/crates/sqlite/src/lib.rs
@@ -616,6 +616,24 @@ impl PersistenceReader for SqlitePersistence {
         self._get_persistence_global(key)
     }
 
+    async fn max_ts(&self) -> anyhow::Result<Option<Timestamp>> {
+        let max_committed_ts = {
+            let connection = &self.inner.lock().connection;
+            // Match TimestampRange::all(), whose Timestamp::MAX upper bound is exclusive.
+            let max_ts: Option<u64> = connection.query_row(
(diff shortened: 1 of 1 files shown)


Andrew-Sem:
Hi! Just checking in on this when you have a chance. The workflows are still waiting for approval; could someone enable them? Happy to make any adjustments needed. Thanks!

PR #547, opened 2026-09-04
Fix hard_delete_tablet_document no-op for Deleting tablets

### LINKED ISSUE #515: Tablets in `Deleting` state are never hard-deleted: cleanup guard checks a table mapping that can never contain them
## Summary

Once a table is dropped (state `Deleting`) and document retention has purged its documents, the tablet's `_tables` metadata row is supposed to be removed by `SystemTableCleanupWorker::cleanup_deleted_tablets` → `TableModel::hard_delete_tablet_document`. That finalization is a permanent no-op, because its existence guard consults the table mapping — which never contains tablets in `Deleting` state. As a result, every backend restart re-discovers the same long-empty tablets and re-runs the (empty) deletion sweep against them, forever.

## Environment

- Self-hosted `convex-local-backend` (Docker), `compiled_revision` `19431ea0dd90bc55ae58dbbd06d9aa045f97336f` (2026-07-23)
- The relevant code is unchanged on current `main`.

## Observed behavior

Our deployment replaced its full table set in late March (snapshot import). The documents in those dropped tablets were purged months ago, yet every restart still logs ~130 lines like:

```
INFO database::retention: go_delete_table_documents: Deleting documents in tablet AbvtxFDNegUh7HJzae5QNw deleted at timestamp 1774305310004627943
INFO database::retention: go_delete_table_documents: Deleting documents in tablet BPhsWL8NXkYdX37qRExQNw deleted at timestamp 1774305437844085164
... (~130 tablets, ~1.5–2 s apart, until)
INFO database::retention: No tablets left to delete
```

The per-tablet pacing is just the retention rate limiter (`check_n(DELETE_TABLET_CHUNK_SIZE = 256)` against `DOCUMENT_RETENTION_RATE_LIMIT = 256/s`) paying one full chunk per **empty** tablet; `rows_deleted` is 0, but that is only visible at DEBUG level. The INFO line ("Deleting documents … deleted at timestamp <4 months ago>") therefore reads as if retention had been stuck for months. The `_tables` rows stay in `Deleting` state indefinitely and the identical sweep repeats on the next boot.

## Root cause

1. When the sweep finds a tablet empty, it sends it to the cleanup worker: [retention.rs#L1291-L1354](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/database/src/retention.rs#L1291-L1354) (INFO log at L1310 is emitted unconditionally, before checking whether anything is left to delete).
2. The worker calls `hard_delete_tablet_document`: [system_table_cleanup/mod.rs#L177-L213](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/application/src/system_table_cleanup/mod.rs#L177-L213)
3. That function only acts `if self.tx.table_mapping().tablet_id_exists(tablet_id)`: [bootstrap_model/table.rs#L265-L279](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/database/src/bootstrap_model/table.rs#L265-L279)
4. But a tablet in `Deleting` state is never present in the table mapping:
   - the live drop removes it from the mapping in the same commit that sets `Deleting`: [table_registry.rs#L310-L329](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/database/src/table_registry.rs#L310-L329)
   - bootstrap after a restart skips `Deleting` tablets entirely: [database.rs#L461-L483](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/database/src/database.rs#L461-L483)
   - `tablet_id_exists` is a plain `tablet_to_table.contains_key`: [value/table_mapping.rs#L179-L181](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/value/src/table_mapping.rs#L179-L181)

So the guard is always false, the `_tables` document is never deleted, and `tablets_to_delete()` ([retention.rs#L467-L485](https://github.com/get-convex/convex-backend/blob/19431ea0dd90bc55ae58dbbd06d9aa045f97336f/crates/database/src/retention.rs#L467-L485)) re-collects the same tablets at every boot. The `TableUpdateMode::HardDelete` path in the table registry appears to be unreachable via this flow.

## Impact

Cosmetic but confusing: minutes of misleading INFO spam on every start (operators reasonably conclude retention has been broken for months), a bit of retention rate-limiter budget burned per boot, and `_tables` tombstone rows that accumulate forever — one per dropped table, e.g. after every `snapshot import --replace` or schema table drop.

## Repro

1. Create a table with a few documents, then drop it (or run a snapshot import that replaces tables).
2. Wait past `DOCUMENT_RETENTION_DELAY`, then restart the backend so the tablet sweep runs and empties the tablet.
3. Restart again: the same tablet is re-announced with "Deleting documents in tablet …", and its `_tables` row is still in `Deleting` state. This repeats on every subsequent restart.

## Suggested fix

In `hard_delete_tablet_document`, don't gate on `table_mapping().tablet_id_exists` (deleting tablets are intentionally unmapped). Either look up the `_tables` document directly (`table_resolved_doc_id` + `tx.get`, treating "not found" as already-deleted to keep idempotency) or consult `tablet_states` (which does track `Deleting`), then delete the metadata document as today.

---

Found while auditing restart logs of a production self-hosted deployment; analysis assisted by Claude. Happy to provide more logs if useful.

rit3sh-x:
Fixes #515.

`hard_delete_tablet_document` gated its delete on `table_mapping().tablet_id_exists(tablet_id)`. A tablet in `Deleting` state is removed from the mapping in the same commit that starts the drop and never re-added on bootstrap, so the guard is always false at the sole call site -> the `_tables` doc is never hard-deleted -> retention re-discovers the same empty tablets on every restart forever, spamming misleading INFO logs and leaving tombstone rows.

Fix resolves the `_tables` doc directly; missing doc = already finalized = idempotent no-op.

Added regression tests under `--features testing`:
- hard_delete_removes_deleting_tablet_metadata
- hard_delete_refuses_non_deleting_tablet

### DIFF
--- a/crates/database/src/bootstrap_model/table.rs
+++ b/crates/database/src/bootstrap_model/table.rs
@@ -262,19 +262,34 @@ impl<'a, RT: Runtime> TableModel<'a, RT> {
         Ok(table_metadata.number)
     }
 
-    /// This method should only be called after all documents in the tablet have
-    /// been deleted by retention
+    /// Removes the `_tables` metadata document for a tablet whose documents
+    /// document retention has finished deleting, finalizing the drop.
+    ///
+    /// This must not gate on the table mapping. A tablet in `Deleting` state is
+    /// removed from the mapping in the commit that starts the drop and is never
+    /// re-added (bootstrap skips `Deleting` tablets), so
+    /// `table_mapping().tablet_id_exists(tablet_id)` is always false here and
+    /// gating on it makes the whole method a no-op. Resolve the `_tables`
+    /// document directly instead. A missing document means the tablet was
+    /// already finalized -- the cleanup worker can be handed the same tablet
+    /// more than once (e.g. re-swept by retention after a restart) -- which is a
+    /// no-op so the call stays idempotent.
+    ///
+    /// Only call this once document retention has emptied the tablet.
     pub async fn hard_delete_tablet_document(&mut self, tablet_id: TabletId) -> anyhow::Result<()> {
-        if self.tx.table_mapping().tablet_id_exists(tablet_id) {
-            let doc = self.get_table_metadata(tablet_id).await?;
-            anyhow::ensure!(
-                doc.state == TableState::Deleting,
-                "Cannot delete a tablet that is not in deleting state"
-            );
-            SystemMetadataModel::new(self.tx, TableNamespace::Global)
-                .delete(doc.id())
-                .await?;
-        }
+        let table_doc_id = self.tx.bootstrap_tables().table_resolved_doc_id(tablet_id);
+        let Some(doc) = self.tx.get(table_doc_id).await? else {
+            return Ok(());
+        };
+        let doc: ParsedDocument<TableMetadata> = doc.parse()?;
+        anyhow::ensure!(
+            doc.state == TableState::Deleting,
+            "Cannot hard-delete tablet {tablet_id} in state {:?}, expected Deleting",
+            doc.state,
+        );
+        SystemMetadataModel::new(self.tx, TableNamespace::Global)
+            .delete(doc.id())
+            .await?;
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


emmaling27:
thanks for the PR! we're working on merging it :)

emmaling27:
merged in https://github.com/get-convex/convex-backend/commit/daf4240bb2a70a6084f1a894729d55591cc1ae0f

PR #552, opened 2026-09-08
node-executor: don't cache source packages missing files on disk

### LINKED ISSUE #519: Self-hosted: empty source-package extraction is cached as success, permanently breaking all Node actions (Cannot find module .../modules/X.js)
## Summary

On a self-hosted backend, **every** Node action (`"use node"`) failed permanently with `Cannot find module …/modules/<name>.js`, because the source package directory was created but left empty. The failure never healed on its own: the in-memory package cache treats the package as available, so no re-download is ever attempted. Only restarting the container fixed it — no re-push, no code change.

The interesting part is not that an extraction failed, but that a *failed* extraction is indistinguishable from a successful one on every subsequent invocation.

## Environment

| | |
|---|---|
| Backend image | `ghcr.io/get-convex/convex-backend@sha256:467964cc6af57ba3e757e3e6cb1fa09a1c577803a19f03f0f42c9c4b134b070c` |
| Deployment | Docker Compose (via Coolify), Postgres 17, local file storage on a volume at `/convex/data` |
| CLI | `convex` 1.42.3 |
| `convex.json` | `node.externalPackages: ["@hyzyla/pdfium", "unpdf", "sharp", "@react-pdf/renderer"]` |

(`/version` returns `"unknown"` on this image, hence the digest.)

## Symptom

Two unrelated Node actions, both failing identically — same temp dir, same source package UUID:

```
Cannot find module '/convex/data/tmp/.tmp6Aj2GR/source/8ff89851-9892-458a-bbb3-0d70a87651d6/modules/extraktion.js'
  imported from /convex/data/tmp/.tmp6Aj2GR/local.cjs

Cannot find module '/convex/data/tmp/.tmp6Aj2GR/source/8ff89851-9892-458a-bbb3-0d70a87651d6/modules/devRequesty.js'
  imported from /convex/data/tmp/.tmp6Aj2GR/local.cjs
```

Queries and mutations were unaffected. `npx convex deploy` had completed successfully earlier, and `npx convex function-spec` listed both modules including inferred return types — so analyze had imported them at push time.

## On-disk state before the restart

```
/convex/data/tmp/.tmp6Aj2GR/
  srwxr-xr-x  .executor.sock
  drwx--r--   external_deps/     <- populated, ~74 MB
  -rw-r--r--  local.cjs          2580500
  -rw-r--r--  local.cjs.map      4516004
  drwx--r--   source/
              └── 8ff89851-9892-458a-bbb3-0d70a87651d6/   <- link count 2: EMPTY, no modules/
```

Notable:

* **`external_deps/` extracted fine, the source package did not.** Both are downloaded in the same `Promise.all`, so this was not a general I/O or network failure.
* **Not a disk space problem** — `df` reported 37G used of 1008G (4%).
* **No retry ever happened.** The mtime of `source/` stayed at the container start time (12:01) even though Node actions were invoked repeatedly until ~17:35. A plain download failure would have been retried on each invocation and shown up in the logs; this did not.

## Why it is permanent

In `npm-packages/node-executor/src/source_package.ts`, `maybeDownloadAndLinkPackages` returns on a cache hit without checking whether the files are still on disk:

```js
const local = availableSourcePackages.get(sourcePackage.key);
if (local !== undefined) {
  return local;
}
```

So once a key is registered as available — whether or not `modules/` actually materialized — every later invocation short-circuits to a broken directory and fails at `await import(...)`. This turns what should be a transient, self-healing error into a permanent outage of all Node actions until the process is restarted.

I have not been able to determine *why* the extraction produced nothing in the first place; my guess is a race at container startup, but that part is unconfirmed.

## Suggested fix

Either of these would make the failure self-healing:

1. Verify the extracted module set against `metadata.json` **on disk** (file existence, not just the expected-vs-metadata name list) before inserting into `availableSourcePackages`.
2. On `ERR_MODULE_NOT_FOUND` at import time, evict the cache entry and re-download once before surfacing the error.

## Workaround

`docker restart <backend-container>` — new process, empty cache, forced re-download.

## What I could not capture

The logs from the window in which the extraction failed were no longer available by the time the cause was understood (the container had been restarted since). If this recurs I will grab `docker logs` before restarting and add them here. Happy to provide any further detail from the environment in the meantime.

siddiqueirshad:
## Summary

Fixes #519.

On self-hosted backends, a failed or empty source-package extract could be treated as success and then pinned in `availableSourcePackages` forever. Later Node action invocations hit the cache and never re-downloaded, so every `"use node"` action failed with `Cannot find module …/modules/X.js` until the process was restarted.

Two independent gaps caused that:

1. **Extract validation used zip entry names, not the filesystem.** `processSourcePackageStream` compared `metadata.json` to `zip.getEntries()`, then cached the package even if `extractAllTo` wrote nothing (or only `metadata.json`).
2. **Settled cache hits never rechecked disk.** Once a key was in `availableSourcePackages`, `getOrStartDownload` returned it even if the directory had since been emptied.

## Change

- After unzip, require `metadata.json` and every listed module file to exist on disk before the package is considered ready. Failures stay uncached (existing `getOrStartDownload` catch path).
- On a settled source-package cache hit, revalidate those files. If they are gone, drop the entry and download again. In-flight downloads for the same key still share one extract.
- Skip incomplete Lambda prebuild directories instead of caching them.

## Test plan

- [x] `npx vitest run src/source_package.test.ts` in `npm-packages/node-executor` (3 passing):
  - extract writes module files before caching
  - a metadata-only extract throws and is not cached
  - deleting a cached package directory causes the next call to re-download

By submitting this pull request, I confirm that Convex can use, modify, copy, and redistribute the contribution, under the terms of its choice.

### DIFF
--- a/npm-packages/node-executor/src/source_package.test.ts
+++ b/npm-packages/node-executor/src/source_package.test.ts
@@ -1,0 +1,183 @@
+import { createHash } from "node:crypto";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { pathToFileURL } from "node:url";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const { partialExtract } = vi.hoisted(() => ({
+  partialExtract: { value: false },
+}));
+
+vi.mock("adm-zip", async () => {
+  const { default: ActualAdmZip } =
+    await vi.importActual<typeof import("adm-zip")>("adm-zip");
+  return {
+    default: function MockAdmZip(
+      ...args: ConstructorParameters<typeof ActualAdmZip>
+    ) {
+      const zip = new ActualAdmZip(...args);
+      const originalExtract = zip.extractAllTo.bind(zip);
+      zip.extractAllTo = (outputDir: string, overwrite?: boolean) => {
+        if (!partialExtract.value) {
+          return originalExtract(outputDir, overwrite);
+        }
+        const metadata = zip.getEntry("metadata.json");
+        if (metadata === null) {
+          throw new Error("zip is missing metadata.json");
+        }
+        fs.mkdirSync(outputDir, { recursive: true });
+        fs.writeFileSync(
+          path.join(outputDir, "metadata.json"),
+          metadata.getData(),
+        );
+      };
+      return zip;
+    },
+  };
+});
+
+import AdmZip from "adm-zip";
+import {
+  availableSourcePackages,
+  maybeDownloadAndLinkPackages,
+  PackageRefcounts,
+  type SourcePackage,
+} from "./source_package";
+
+const MODULE_PATH = "actions/hello.js";
+const MODULE_SOURCE =
+  'export default { isAction: true, invokeAction: async () => "{}" };\n';
+
+function writeSourceZip(
+  zipPath: string,
+  modulePath: string,
+  source: string,
+): { sha256: string } {
+  const zip = new AdmZip();
+  const metadata = {
+    modulePaths: [modulePath],
+    moduleEnvironments: [[modulePath, "node"]],
+  };
+  zip.addFile("metadata.json", Buffer.from(JSON.stringify(metadata)));
+  zip.addFile(`modules/${modulePath}`, Buffer.from(source));
+  const zipBuffer = zip.toBuffer();
+  fs.writeFileSync(zipPath, zipBuffer);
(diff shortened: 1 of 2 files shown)


siddiqueirshad:
Hi @jordanhunt22, when you get a chance, could you please take a look at this PR? Thanks!

PR #556, opened 2026-09-11
Report invalid header/cookie values instead of a generic 500

rit3sh-x:
When a cookie value contains a raw, non-percent-encoded non-ASCII character (e.g. ä, ö, ü), the byte carrying it doesn't fit in the u8 range `byte_string_to_header` requires when converting a JS-side header value back into an `http::HeaderValue`. That conversion failed with a bare `anyhow!("invalid char for header...")` and no header name attached, since it wasn't tagged with `ErrorMetadata`, it fell through to the fully generic `INTERNAL_SERVER_ERROR_MSG`.

### DIFF
--- a/crates/isolate/src/http.rs
+++ b/crates/isolate/src/http.rs
@@ -11,6 +11,7 @@ use common::{
     },
     sync::spsc,
 };
+use errors::ErrorMetadata;
 use futures::{
     stream::BoxStream,
     FutureExt,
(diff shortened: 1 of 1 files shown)


rit3sh-x:
@emmaling27 could you review it?

emmaling27:
Thanks for the contribution! merged in bb3e09771582f940da3ec8f9c557756167c99b45

rit3sh-x:
Hey @emmaling27, just wanted to clarify: is this repo a public mirror of the main Convex backend repository? I noticed that my PRs are being closed after the changes are incorporated into commits, so I was wondering if that’s the expected workflow for contributions here.

ianmacartney:
Yes we apply the changes to the private repo then the changes are reflected back here, in a way that doesn't register as a merge of the original PR. You're doing it correctly. Ideally we could mark it as merged so it would feel more gratifying / visibility of contribution. You're listed as a Co-author on [the commit](https://github.com/get-convex/convex-backend/commit/bb3e09771582f940da3ec8f9c557756167c99b45).

